PR #110, opened 2026-03-02
Staging performance improvements and developer tooling

luismetzger:
This PR addresses the performance issues we were seeing on the allocations dashboard in staging, along with some quality-of-life improvements for developers working with the staging environment.

## What was happening

The `/allocations/` page was taking a long time to load, sometimes timing out entirely. Looking at CloudWatch metrics, the ECS container was hitting 99% CPU utilization during page loads. The culprit was a combination of:

1. Server-side chart generation using Matplotlib (CPU-intensive, especially under load)
2. An N+1 query pattern where we were calling `get_facility_overview_data()` once per resource instead of batching
3. The container was undersized for the workload (512 CPU units, 1GB memory)

## Performance fixes

**ECS resource bump** - Doubled the container resources to 1024 CPU units and 2GB memory. This gives us more headroom and should prevent the CPU saturation we were seeing.

**Flask-Caching** - Added response caching with a 5-minute TTL for the allocations index and projects fragment routes. The allocation data doesn't change frequently, so caching makes sense here. The cache is disabled in test mode to keep tests deterministic.

**Cached Matplotlib chart generation** - Instead of moving to client-side Chart.js, we kept the familiar server-side Matplotlib rendering and added `functools.lru_cache` to all 5 chart generation functions. This caches the generated SVG strings so repeated requests with the same data skip the expensive rendering entirely. The public API is unchanged -- callers pass normal Python objects and the caching is handled internally via hashability wrappers. This gives HPCD developers a predictable interface to Matplotlib (and eventually Plotly) while eliminating the CPU spikes.

**Query batching** - Replaced the per-resource `get_facility_overview_data()` calls with a new `get_all_facility_overviews()` function that fetches everything in a single call to `get_allocation_summary()`. Fewer round trips to the database.

## Developer tooling

While working on this, I also improved the developer experience for working with the staging environment:

**Prerequisite checking library** - Added `scripts/lib/prereqs.sh` with helper functions that all the infra scripts now use. When you run a script and are missing a dependency (like `mysql-client`), it will automatically install it via Homebrew on macOS instead of just printing an error. This should save some frustration for developers setting up their environment.

**New scripts:**

- `db-creds-staging.sh` - Fetches the staging DB credentials from AWS SSM. Useful for sharing credentials with team members who need database access but don't have AWS CLI configured.
- `switch_to_staging_db.sh` - Configures your local `.env` to point at the staging database (requires VPN).

**Updated scripts** - All the existing infra scripts (`query-staging-db.sh`, `deploy-staging.sh`, `ssh-staging.sh`, etc.) now use the prereqs library for consistent dependency checking.

## Infrastructure changes

- Enabled deletion protection on the RDS instance (should have been on from the start)
- Added `FLASK_SECRET_KEY` to the ECS task definition secrets
- Changed database subnet routing to go through the Internet Gateway instead of NAT Gateway, which allows developers on VPN to connect directly to RDS

## CI/CD and pre-commit

- Added `.pre-commit-config.yaml` with trailing whitespace, YAML checks, branch protection, terraform fmt, and GitGuardian (on push)
- Added `ci-staging.yaml` workflow for PRs targeting staging (secret scan, lint, terraform validation, test suite)
- Added pre-deploy validation to `deploy-staging.yaml` (TruffleHog + terraform fmt check)

## Testing

Added `test_allocations_performance.py` with tests covering:

- Caching behavior (cache hits, cache disabled in test mode)
- Query batching (verifying the new function works correctly)
- Matplotlib lru_cache behavior (cache hits/misses, observability, test isolation)
- Route behavior (happy path, error handling, edge cases)

617 tests pass across the full suite.

## How to test

1. Deploy to staging and hit the `/allocations/` page - should load noticeably faster
2. Check CloudWatch metrics - CPU should stay well below 100% now
3. Try running `./scripts/infra/query-staging-db.sh` without mysql-client installed - should auto-install

Let me know if you have any questions or want me to walk through any of the changes.


benkirk:
@luismetzger I like nearly all of this; but am wondering if we can undo 

> Client-side charts - Moved the pie chart rendering from server-side Matplotlib to client-side Chart.js. This offloads the rendering work to the browser and eliminates the CPU spike from generating SVGs on every request. The charts look essentially the same, just rendered differently.

and purse a less dramatic change first.  Heres the idea:

- HPCD developers will be well familiar with matplotlib,
- I understand the implications of server-side chart generation, BUT
- let's first try caching the matplotlib functions.  Allocations information and metrics-related info coud be cached on two different timescales.

I think the built-in Python `functools` `lru_cache` might work fine here.

Basically, I'm thinking we can solve the chart generation issue with caching alone and allowing devs to have a predictable interface to matplotlib (or additionally plotly) and keep server-side generation.

luismetzger:
@benkirk - Reverted the Chart.js migration and went with `functools.lru_cache` on the chart generation functions instead.

The approach: each of the 5 Matplotlib functions in `charts.py` now has an `lru_cache` wrapper that converts the list-of-dicts input to a hashable key. First request pays the rendering cost, subsequent calls with the same data return the cached SVG string immediately. The public API is unchanged -- callers still pass normal Python objects.

This stacks with the Flask-Caching route cache (5-min TTL) as a second layer. If the route cache expires, the chart SVGs are still cached in-process.

The Chart.js CDN and `pie-charts.js` have been removed. Tests updated accordingly -- replaced the Chart.js data contract tests with cache behavior tests (hits, misses, observability). All 617 tests pass.

Commit: ad655f5

PR #316, opened 2026-06-15
Public-exposure tooling: weblog audit + shared script lib + R1/R2 hardening

benkirk:
## Summary

Operational tooling and hardening for the now-public `samuel` deployment, built and **live-tested on CIRRUS (`nwc1`)** via manual `workflow_dispatch` deploys. Three parts:

1. **`scripts/cirrus_weblog_audit.sh`** — a read-only, on-demand traffic / rate-limit / abuse audit, companion to `cirrus_healthcheck.sh`. Harvests the webapp's stdout (+ the Redis `ratelimit:events` set) and reports traffic & status mix, top talkers, scanner/probe signatures, 429 offenders, auth/CSRF failures, and edge-rate-limit headroom — with a PASS/WARN/FAIL verdict (exit 0/1/2).
2. **`scripts/lib/{common,cirrus_common}.sh`** — a two-layer shared shell library; `cirrus_healthcheck.sh` and `zap_probe_docker.sh` migrated onto it (behavior-preserving) to stop duplicating colors/log helpers/arg-parsing/constants.
3. **R1 + R2 public-exposure hardening** (helm + app), described below.

## Key finding: the real client IP is lost at the ingress (R2)

Verified live with an **external, off-VPN request** (from Anthropic infra): the `nginx-external` controller hands the app `X-Forwarded-For: 127.0.0.1` for **every** client. Two consequences:

- **R2 (real client IP)** cannot be fixed in this repo — there is no client IP in the headers to recover. `ProxyFix` is wired and now env-tunable (`PROXYFIX_X_FOR`, default **1**, safe), and gunicorn now logs the `xff="…"` chain so we can verify a fix later. Nothing to tune until the controller forwards real IPs.
- **R1 (edge rate limiting)** therefore keys every client into **one global bucket**, not per-IP. Set as a deliberately **high global flood-backstop** (`limit-rps: 100`, `limit-connections: 200`) so legitimate aggregate traffic won't trip 503s; `rps: 0` disables it. The audit's section 1 shows live `limit-rps` vs observed peak load so we'll know if it needs raising (actual edge 503s happen at the controller and aren't in app logs — noted in-tool).

## ⛳ Forward work — ingress-NGINX controller (needs the `nwc1` platform team)

This is the keystone that fixes R2 **and** makes R1 per-IP meaningful, in one cluster-side change (out of scope for this repo — controller ConfigMap, affects all ingresses on that controller):

- Enable `use-forwarded-headers: "true"` + `compute-full-forwarded-for: "true"` **if** a trusted L7 proxy/LB already sets `X-Forwarded-For`; **or**
- `use-proxy-protocol: "true"` if the upstream LB (`llb-128.117.41.126`) speaks PROXY protocol instead.
- Set `proxy-real-ip-cidr` to the LB range so client-supplied XFF can't be spoofed.

Once in place: re-run `cirrus_weblog_audit.sh` to confirm real client IPs flow, then bump `webapp.proxyFixForwardedHops` to the now-visible hop count and re-shape `webapp.ingress.rateLimit` back to sane **per-IP** values.

## Testing

- `bash -n` clean on all scripts/libs; `helm template` renders R1 annotations and the `rps: 0` off-switch; `cirrus_healthcheck.sh` migration verified live (26 PASS / 2 WARN / 0 FAIL, unrelated transient rollout warns).
- `cirrus_weblog_audit.sh` exercised live with controlled traffic: status mix, probe detection, scanner-vs-CLI UA split, Redis `ZREVRANGE` enrichment, the "no public client IPs" detector, and the edge-headroom indicator all confirmed firing. Health-probe paths and `/static/` assets excluded from analysis.
- `shellcheck` not run (not installed locally).

## Deploy notes

- CIRRUS deploys from `main`; this branch was tested via manual `workflow_dispatch`.
- R2 plumbing is safe/inert at `PROXYFIX_X_FOR=1`.

### DIFF
--- a/containers/webapp/gunicorn_config.py
+++ b/containers/webapp/gunicorn_config.py
@@ -47,10 +47,22 @@ keepalive = 5
 preload_app = True   # Load app before forking → faster startup, shared memory
 daemon = False       # Stay in foreground (required for containers)
 
+# Trust X-Forwarded-* from the ingress. gunicorn is only reachable via the
+# in-cluster ingress (ClusterIP service, never published directly), so trusting
+# forwarded headers from any peer is the standard k8s setting and lets it honor
+# X-Forwarded-Proto for correct https scheme detection. Real client-IP recovery
+# for the app is handled by ProxyFix (see webapp/run.py); this is the gunicorn
+# side of the same story.
+forwarded_allow_ips = os.environ.get('FORWARDED_ALLOW_IPS', '*')
+
 # Logging — stdout/stderr by default (Docker captures them)
 accesslog = os.environ.get('GUNICORN_ACCESS_LOG', '-')
 errorlog  = os.environ.get('GUNICORN_ERROR_LOG',  '-')
 loglevel  = os.environ.get('GUNICORN_LOG_LEVEL',  'info')
-access_log_format = '%(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s" %(D)sµs'
+# %(h)s is the raw socket peer (the in-cluster proxy), so it is NOT the client.
+# The trailing xff="…" logs the full X-Forwarded-For chain: its leftmost entry
+# is the real client, and counting the entries tells us the correct
+# PROXYFIX_X_FOR hop depth (see webapp/run.py).
+access_log_format = '%(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s" %(D)sµs xff="%({x-forwarded-for}i)s"'
 
 proc_name = 'sam-webapp'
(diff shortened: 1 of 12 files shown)


benkirk:
### ✅ Post-redeploy verification (nwc1, live)

Branch deployed via manual `workflow_dispatch`:

- **R1 live:** ingress `limit-rps=100`, `limit-connections=200`, `limit-burst-multiplier=5`.
- **R2:** `PROXYFIX_X_FOR=1` (safe default; nothing to recover until the controller forwards real IPs).
- **Audit headroom indicator** now reads `Edge limit-rps: 100/s (global) · observed peak: 5/s` → PASS.
- **Audit section 2** still correctly flags `no public client IPs … real client IP not propagated by the ingress` (R2 confirmed controller-side).
- Status collectors (`python-requests` → `/api/v1/status/*`) correctly classified as benign automation.

Open item unchanged: the ingress-NGINX controller change (forward work section above) remains the keystone for real per-IP limiting + client-IP logging.

PR #375, opened 2026-07-26
Webapp quick wins: dead code, stale docs, charts lifecycle, Flask-Admin registry, ownership fixes

benkirk:
## Summary

Track 2 ("peripheral quick wins") of the webapp boilerplate-reduction plan
(`docs/plans/WEBAPP_OO_REFACTOR.md`, committed here as the first commit).
**Net −1,105 lines** across 8 commits, all behavior-preserving except two
deliberate, flagged improvements (below). Track 1 (the form-layer OO
refactor) follows as a separate stacked PR.

### Commits (in order)
1. **Plan doc** — `docs/plans/WEBAPP_OO_REFACTOR.md` (investigation summary + progress checklist).
2. **Dead code** (−164): `success_response`/`error_response` (0 callers; documented an envelope that contradicts the deployed contract), `require_any_permission`/`require_role` + their orphaned predicates, empty `webapp/clients/` package.
3. **Stale docs** (−702): delete `DESIGN.md` + `REFACTORING_PLAN.md` (Bootstrap-4-era, flagged by the 2026-05 NRIT review; git history preserves them), refresh `README.md` (stack, run instructions, tree, group-bundle RBAC), add the function-local import policy, fix `FORMAT_DISPLAY.md` status line.
4. **charts.py** (−13, leak fix): `_fig_to_svg()` closes figures in a `finally` (previously a savefig exception leaked the figure); `_empty_state()` replaces 14 hand-written placeholder divs — rendered markup byte-identical.
5. **Ownership fixes**: `invalidate_queue_cache()` now lives in `api/v1/queue.py` beside the memoized functions it clears (additive touch approved; response bytes unchanged, locked by `tests/api/test_queue_access.py`); deletes the verbatim copy in `resources_routes.py` that imported the API module's private `_queue_data`. New `rbac.allowed_facility_names()` replaces 3 hand-rolled copies.
6. **GET-side consolidation**: the 4 copy-pasted date-range parse blocks in `user/blueprint.py` adopt `parse_input_start_date`/`parse_input_end_date`; shared `sort_link` macro for the transactions/adjustments tables.
7. **Flask-Admin spec-loop** (−323): the 93 generated `class XDefaultAdmin(SAMModelView): pass` declarations + 93 registration lines collapse to a loop — Flask-Admin distinguishes views by endpoint, not class.
[...]

PR #376, opened 2026-07-26
Form-layer OO refactor: HtmxFormHandler lifecycle, CRUD registrar, typeahead factory

benkirk:
Phase 2 of the webapp boilerplate-reduction plan (`docs/plans/WEBAPP_OO_REFACTOR.md`; Phase 1 was #375). Six ordered commits, each independently green on the full suite.

## What this does

**Commit 2.1 — `HtmxFormHandler` lifecycle class** (`webapp/utils/form_handler.py`)
Template-method base expressing the HTMX form flow: `form_input() → load() → clean() → [management_transaction: perform()] → after_commit() → on_success()`, with every error path funneling through `render_errors()`. `handle_htmx_form_post` keeps its exact signature but is now a thin adapter over the same lifecycle (zero churn for its ~35 call sites), plus a new `after_commit=` kwarg and a `modal_triggers()` helper. Audit logging is now enforced by construction — a handler cannot write outside `management_transaction`.

**Commits 2.2–2.3 — declarative CRUD registrar** (`webapp/dashboards/admin/crud.py`)
`CrudSpec` + `register_crud` generate the edit-form/edit/create-form/create/delete quintet for 14 admin entities (Organizations card ×8, facility, panel, allocation-type, resource, resource-type, machine). Endpoints, URL rules, templates, permissions, and not-found messages are identical to the hand-written routes (pinned by 145 characterization tests written *before* the migrations, plus the route-map snapshot). Hard rule in the docstring: an entity needing more than the spec stays bespoke — mnemonic-code, contract-delete, panel-session, and the decommission-by-date deletes all stayed.

**Commit 2.4 — 21 hand-rolled handlers → handler subclasses**
The big one: project-update, the six allocation workflows (add/exchange/allocate-down/renew/extend/edit), linked-elements adds, admin directory modals, access-grid toggle, shell/primary-GID/threshold, add-member, queue-edit, disk-root modals. Routes shrink to entity-load + `Handler(...).handle()`.

**Commit 2.5 — §9 conformance**
New `sam/schemas/forms/operational.py` (wallclock-exemption schemas with legacy-verbatim error messages); the exemption trio's ~280 lines of strptime/float ladders become three handler subclasses sharing one `_resources_with_queues()`. `ChangeProjectAdminForm` + `LinkAllocationParentForm` close out the last unvalidated POSTs.

**Commit 2.6 — `register_typeahead` factory** — 7 of 9 search-as-you-type endpoints; the multi-context user search and facility-scoped project search stay bespoke because their branching is the feature.

## Error-UI unification (deliberate change)

Per the Phase-0 decision, migrated forms move from top-panel flattened error lists to **inline field errors** (`split_errors` → `form_fields.html` macros, completing the #336 follow-up). Templates without per-field macros keep the legacy panel presentation via the `FlattenedFieldErrors` mixin, so nothing loses error visibility. The add-member picker's hidden `username` field reroutes its errors into the visible panel.

## Behavior deltas (small, intentional — called out per commit)

- Panel edit is now schema-validated (was raw `request.form` coercion).
- Queue-create's unknown-resource message drops the `Error creating queue:` prefix; queue cache invalidation fires only on committed success.
- add-allocation / access-grid report FK errors on a separate render from schema errors (previously accumulated).
- save-threshold and access-grid unexpected exceptions re-render with a message instead of 500ing.
- The admin edit-allocation "latent 500" flagged during design turned out to be already fixed by #373 — no change.
- Pre-existing quirk, now fixed (52a1955 + ac46a1b): `htmx_form` reads its `errors` *keyword argument*, not the template context, so templates that didn't pass `errors=errors` silently dropped form-level messages. On the mnemonic-code create form this made **all** validation feedback invisible (its route uses the flattened panel list); on 28 other fragments only the exception-fallback path (`Error creating X: …`) was affected, since inline field errors ride the implicit `field_errors` context. Every `htmx_form` caller now passes `errors=errors` (harmless falsy-undefined on GET renders), the macro docstring states the pass-through as mandatory, and the mnemonic pinning tests now assert the messages render.

## Numbers

- Source: **−729 lines net** (2,247+ / 2,976−) across `src/`; `orgs_routes.py` 1111→509, `facilities_routes.py` 471→224.
- Tests: **+2,654 lines** — 145 CRUD characterization tests (written pre-migration), 22 handler-lifecycle unit tests, 22 HTTP-layer handler tests, and a **route-map parity test** pinning all 239 dashboard `(endpoint, rule, methods)` triples to a checked-in snapshot (unchanged throughout).
- Grep gates: `except ValidationError` in dashboards 31 → 11, every survivor documented (wizards, alert-response routes, outage pair per #337).
- Full suite: **3098 passed**, 30 skipped, 1 xfailed (~95 s); perf query-count baselines unchanged.

## Verification

Playwright smoke complete — [results posted below](https://github.com/benkirk/sam-queries/pull/376#issuecomment-5084402909). All probe groups green (registrar not-found asymmetry, inline field errors, queue-edit cache round-trip in both directions, typeaheads). The smoke caught one real bug — a 500 on the admin exemption-create error path — fixed in d9251f3 with three regression tests; final suite tally is 3101 passed.


benkirk:
## Playwright smoke results (local webdev, branch code verified in container)

Fetch-based probes through a real logged-in session (CSRF via page meta → `X-CSRFToken`), plus a live queue round-trip against the dev DB.

**Registrar quintets (Organizations card)** ✅
- Edit form renders for a real org; missing id → warning div at 200; POST to missing id → 404 (not-found asymmetry preserved).
- Invalid edit (blank name/acronym) → 200 re-render with **inline field errors** (`.is-invalid` + `.invalid-feedback` present), no success trigger.

**Panel edit (new `EditPanelForm`)** ✅ — malformed `active=maybe` re-renders with an error instead of `bool()`-coercing to True (the intended delta from 2.3).

**Handler subclasses** ✅
- `project-update` with a 300-char title → inline field error, no write.
- `add-member` empty POST → hidden-`username` error correctly rerouted into the visible panel.
- Queue edit: negative wallclock → inline error; then a full round-trip — **12 → 13, `GET /api/v1/queue` reflects 13 immediately; revert 13 → 12, API immediately 12** (parsed JSON traversal, not string matching). `after_commit=invalidate_queue_cache` works both directions; DB confirmed restored to 12.00.

**Typeahead factory** ✅ — groups, org FK, parent-project, and adjustment pickers all 200; sub-min-length query returns the empty response.

**One real bug caught and fixed (d9251f3)** 🐛
Empty POST to `/admin/htmx/admin/exemption/create` **500'd**: the admin variant's error-render context called `super().context()`, which reads `self.sam_user` — only set once `clean()` runs, but schema errors fire first. Same probe also showed `queue_id` errors were invisible (the cascading Resource→Queue select is inline, not a form_fields macro — no inline error slot); they now surface in the top panel. Re-probed after the fix: 200 with 'User is required.' inline and 'Queue is required.' in the panel. Three regression tests added; full suite now **3101 passed**.

Environment restored: queue 247 back to 12.00; no other writes performed (all other probes were validation-error paths).

PR #380, opened 2026-07-26
docs: plan for a front-end regression net (docs-only)

benkirk:
**Docs-only.** Adds `docs/plans/FRONTEND_TEST_NET.md`; no code changes. Implementation lands on this same branch, ticking the doc's Progress checklist and updating this PR as each part goes in.

## Why

PR #378 fixed two front-end bugs that no automated test could have caught — a dead edit pencil (missing modal shell) and a scroll overshoot past a card title. The repo has ~3,100 Python tests and zero coverage of the 2,571 lines of webapp JS. This doc is the decision record for what to do about that.

## What it decides

**Rejected — a JS unit harness (vitest/jsdom):**

- ~87% of the JS is htmx-swap/DOM lifecycle glue; the genuinely pure logic is ~150 LOC and **none of it is exported** (every file is an ES5 IIFE, no module system).
- **jsdom zeroes `getBoundingClientRect()`** — so the scroll bug is untestable there *by construction*, i.e. the harness people reach for first would have missed one of the two bugs that prompted this.
- It would be the repo's first node dependency, in a codebase that hand-vendors Bootstrap/htmx/jQuery with SRI pins specifically to keep npm out of the serving path.

**Chosen — two complementary pytest-driven layers:**

| Layer | Catches | Cost |
|---|---|---|
| Python shell-contract guards | dead `data-bs-target` shells (**silent in the browser**) | no new deps, ~0s in the existing suite |
| Playwright console sweep (`e2e/`) | dead `hx-target`, JS exceptions, script-order breaks | one dev dep, ~60–90s on a non-critical workflow |

The two-layer split isn't belt-and-braces for its own sake — it was verified empirically against the live app while writing the plan. htmx 2.0.4 *does* `console.error("htmx:targetError")` on a dangling `hx-target`, so the sweep catches that. But a dangling `data-bs-target` produces **no console output at all** (htmx's `logger` defaults to `null`; `htmx-config.js` listens for `responseError`/`sendError`, not `targetError`). A browser-only net would miss Bootstrap-only affordances.

## Notable design points

- **`e2e/` lives outside `tests/`** — `tests/conftest.py`'s DB safety guard hard-exits unless `SAM_TEST_DB_URL` points at an allowlisted host, and `test-install.yaml` doesn't start `mysql-test`. Keeping browser tests outside `tests/` means the guard stays untouched and no `pytest.ini` marker plumbing is needed. Precedent: `utils/parity/check_legacy_apis.py`.
- **CI goes in `test-install.yaml`**, which already checks out the repo, runs `make docker-up`, and curls `:7050` — so there's no server-startup logic to add, only a browser. The canonical 8-minute `sam-ci-docker` path is untouched.
- **The `actions/cache` step is deliberately deferred** — measure the real cost of `playwright install chromium` before introducing the repo's first cache step.
- A flake budget is written down up front: no `--retries`, auto-waiting only, and flaky routes get fixed or deleted rather than retried.

## CI on this PR

`docs/**` is in `paths-ignore` for both `sam-ci-docker.yaml` and `test-install.yaml`, and `mega-linter.yaml` only triggers on PRs to `main` — so this runs essentially nothing. No `[skip ci]` needed.

### DIFF
--- a/docs/plans/FRONTEND_TEST_NET.md
+++ b/docs/plans/FRONTEND_TEST_NET.md
@@ -1,0 +1,310 @@
+# Front-end Regression Net (and why not a JS unit harness)
+
+**Status: PLANNED — not started.**
+Branch: `frontend_test_net` (docs-only PR first; implementation lands on the
+same branch, updating the PR as it goes). Targets `staging`.
+
+> Deliverable of *this* PR: this document only. `docs/**` is in the
+> `paths-ignore` list for both `sam-ci-docker.yaml` and `test-install.yaml`, and
+> `mega-linter.yaml` only triggers on PRs to `main`/`master` — so a docs-only PR
+> to `staging` runs essentially no CI. No `[skip ci]` needed.
+
+## Progress
+
+- [ ] Step 0 — this doc committed, PR opened against `staging`
+- **Part 1 — Python shell-contract guards (no new dependencies)**
+  - [ ] 1.1 `tests/unit/test_modal_shell_contract.py` (generalizes the
+        `test_modal_shell_pairing.py` added in PR #378)
+  - [ ] 1.2 script-order assertion on `dashboards/base.html`
+  - [ ] 1.3 prove it bites — remove the `allocation_modals.html` include, confirm
+        a named failure, restore
+- **Part 2 — Playwright console sweep**
+  - [ ] 2.1 `e2e/` skeleton: `pytest.ini`, `conftest.py` (base_url, error
+        capture, admin login)
+  - [ ] 2.2 `test_console_sweep.py` — ~16 routes + ~5 declared flows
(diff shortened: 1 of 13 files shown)


benkirk:
Superseded by a fresh branch off current `staging`.

This PR's plan predates #406 (the Node 20 Actions-runtime migration), so its entire CI section was written against workflows that no longer exist — `actions/checkout@v4`, no `dependabot.yml`, no actionlint bar, and `test-install.yaml` as the proposed host. Review also turned up two substantive errors in the plan itself: the sweep's route list invented `/status/reservations` (never a route) while omitting seven real pages, and the rate limiter's global 200/min `default_limits` would 429 the sweep.

The revised plan and its implementation land on `ci_playwright`, cut from current `staging`. New PR to follow.

PR #385, opened 2026-07-28
Job History card: time window, empty-band suppression, identifier casing; status Events tab

benkirk:
Four related UI changes, in one commit each — three on the job-history card, one on the status page tab strip.

---

## 1. Selectable time window (`289acde`)

The card's lookback was fixed at 90 days by convention — three inline literals, invisible and unchangeable from the card itself. This adds a **30d / 60d / 90d / 1 yr** pill group to the tab strip that applies to **all six panels at once**, on all three hosts (My Jobs, Status → Job History, project resource-details), with the selection **persisting across machine subtabs and page reloads**.

Because each panel bakes its window into its own `hx-get` URL at render time, a pill re-renders the card *shell* — three new query-free routes (`/<projcode>/card`, `/machine/<machine>/card`, `/user/<machine>/card`) that mirror each mode's existing gating and do no plugin work, so the panels re-fetch lazily as they're shown.

`?days=` is whitelisted against `JOBS_WINDOW_CHOICES` (derived from the same tuple the pills render from, exposed as the `jobs_window_pills` Jinja global) and outranks any `start`/`end` already in the URL — the persistence layer can only append to a request, never rewrite the path. It normalizes back to a plain `start=` in `_roundtrip_params`, so in-panel pills, bar drills and explorer links keep speaking `start`/`end` and nothing downstream learns a second vocabulary. The explorer honours `?days=` too, which is what the card's "Open full view" link hands over.

Persistence rides the existing `data-chart-persist-*` contract under key `days`. Per-machine subtabs share a persist id and stay in lockstep; the window survives a reload. **Resource-details deliberately opts out** — its pills last the visit only, so a stored window can't quietly shadow that page's own date range.

**Two things worth a reviewer's eye:**

- **The card wrapper carries no `hx-target`/`hx-swap`.** htmx inherits both, and an earlier revision with `hx-target="this"` captured every descendant request that didn't name its own target — a By Project bucket drill swapped the whole card away (caught in browser smoke, not by tests). The sibling fan-out goes through `htmx.ajax()` instead, and `test_card_wrapper_declares_no_inheritable_hx_target` pins it.
- **The pill click saves to localStorage synchronously**, before htmx issues any request — the existing `afterSettle` save lands too late for the hand-off to sibling cards.

## 2. Job Sizes hides a leading empty band (`3ef8dd7`)

The 0 band can never fill for Nodes / CPUs — every job uses at least one of each — so it rendered as a permanent empty row plus a gap at the left of the chart. GPUs is the exception worth keeping: there the 0 band holds the CPU-only jobs (933K over 90 days on Derecho), so it survives on its own merit.

`_trim_leading_empty_bands()` sits in the shared histogram renderer, so Wait Times and Durations get the same treatment if their first band ever comes back empty. Two details:

- Emptiness is judged on `job_count` alone, never the displayed metric, so flipping Jobs / CPU-hours / GPU-hours can't shift the axis under the viewer. Zeros *inside* a distribution still render — that's what keeps the x-axis stable across filter changes.
- The trim runs before both the chart and the drill list, since `#jh-bar-<i>` and the table's `data-jh-bucket` are both positions in the same bucket vector. It returns a shallow copy — the envelope is a shared cache entry.

## 3. Identifiers in a button render in their stored case (`66d0902`)

`.btn` carries `text-transform: uppercase`, which is right for button labels but wrong for the entity quick-view links: every username came out shouting and, worse, indistinguishable at a glance from a projcode. Case *is* the convention that tells them apart (`SCSG0001` vs `benkirk`), and a username also has to match what you type at a shell prompt.

CSS-only. Nothing is forced to a case — each identifier renders as stored, so projcodes stay upper and usernames stay lower without the templates having to know which they're holding. Covers both spellings in use: `<code>` in the By User / By Project tables, `.font-monospace` in the histogram band drill-downs.

## 4. Status page: Reservations → Events, moved rightmost (`bacdaca`)

The status tab strip's **Reservations** tab is now **Events**, and sits at the far right of the strip, after the gated tabs.

- Route renamed `/status/reservations` → `/status/events`, endpoint `status_dashboard.reservations` → `.events`, page template → `events_page.html`. The rename is user-facing only — the backing `resource_reservations` rows, the query layer, and the collector ingest API keep the reservation vocabulary.
- Ordered last in both `base_status.html` (tab strip) and `nav.py` (navbar dropdown + mobile offcanvas). The two tabs ahead of it are gated on `VIEW_ALL_FILESYSTEM_DATA` / `VIEW_ALL_JOB_DATA` and drop out of the strip entirely when hidden, so **ordering it last is exactly what makes it rightmost for every audience** — a comment at each list says so, since a future tab appended in the obvious place would break it.
- Data gating and the count in the label are unchanged: the tab still shows only when there are upcoming reservations or a configured calendar embed.

No redirect from the old URL — say the word if bookmarks matter and it's a three-line addition.

---

## Test Results

`3366 passed, 30 skipped, 1 xfailed` (full suite, ~2 min).

44 new tests. Job-history card (38): `?days=` whitelist and precedence, roundtrip normalization, explorer default, the card routes per mode (auth, 404, 400 on unsafe element ids, baked window, pill state, persist opt-in, explorer hand-off, inherited-attribute regression), and band trimming (trim/keep/all-empty/metric-independence plus bar↔row index alignment). Status tab (6): `TestEventsTab` pins the position at both ends of the permission range — staff sees six tabs, a non-admin sees four, Events is last in both — plus page render, data gating, and the label; `test_nav.py` pins the navbar item's position.

Browser-verified throughout: pill click re-renders and reloads the visible panel with the saved tab restored, sibling machine subtabs follow, a reload restores pills + explorer link + data window, My Jobs and Status stay independent, resource-details stores nothing; Nodes starts at the `1` band with bar 0 drilling `min_nodes=1` while GPUs keeps its 0 band; all four identifier surfaces render in the right case; and the Events tab renders rightmost both logged out (4 tabs) and as an admin (6 tabs).

Note on the route-map parity snapshot: the job-history card routes needed no regen (it pins the dashboard blueprints, and those routes live on the `jobs` blueprint), but the status rename does change it — that diff is the endpoint/rule rename and nothing else.

### DIFF
--- a/src/webapp/config.py
+++ b/src/webapp/config.py
@@ -74,7 +74,7 @@ class SAMWebappConfig(SAMConfig):
     LOG_LEVEL = os.getenv('LOG_LEVEL', 'INFO')
     LOG_FILE  = os.getenv('LOG_FILE', '')       # empty = console only
 
-    # Google Calendar embed URL (public calendar shown on reservations tab; empty = hidden)
+    # Google Calendar embed URL (public calendar shown on the Events tab; empty = hidden)
     GOOGLE_CALENDAR_EMBED_URL = os.getenv('GOOGLE_CALENDAR_EMBED_URL', '')
 
     # Status collectors tick every ~5 minutes; a snapshot older than this many
--- a/src/webapp/dashboards/status/blueprint.py
+++ b/src/webapp/dashboards/status/blueprint.py
@@ -226,15 +226,20 @@ def job_history():
     mode — the card's fragment routes are themselves VIEW_ALL_JOB_DATA-
     gated, so this page gate is the first line, not the only one.
 
-    ``jobs_window_start`` bounds every card tab to the last 90 days by
-    default: an unbounded machine-wide aggregation measures ~200 s
-    against the plugin PG vs ~0.6 s per month-window. The explorer's
-    date fields let an operator widen deliberately.
+    ``jobs_window_start`` bounds every card tab to the default window: an
+    unbounded machine-wide aggregation measures ~200 s against the plugin
+    PG vs ~0.6 s per month-window. ``jobs_window_days`` tells the card
+    which period pill to paint active; the pills (and the explorer's date
+    fields) let an operator widen deliberately.
     """
-    from datetime import date as _date
+    from webapp.jobs.service import (
+        DEFAULT_JOBS_WINDOW_DAYS,
+        default_jobs_window_start,
+    )
     return render_template(
         'dashboards/status/job_history_page.html',
-        jobs_window_start=(_date.today() - timedelta(days=90)).isoformat(),
+        jobs_window_start=default_jobs_window_start(),
+        jobs_window_days=DEFAULT_JOBS_WINDOW_DAYS,
         **_page_context(db.session),
     )
 
--- a/src/webapp/dashboards/status/blueprint.py
+++ b/src/webapp/dashboards/status/blueprint.py
@@ -69,7 +69,7 @@ def _page_context(session):
         outages=status_queries.get_active_outages(session),
         # Needed on every page: derecho/casper render per-system
         # reservation cards, and the tab strip shows/hides the
-        # Reservations tab based on it.
+        # Events tab (and its count) based on it.
         reservations=status_queries.get_upcoming_reservations(session),
         google_calendar_embed_url=current_app.config.get('GOOGLE_CALENDAR_EMBED_URL', ''),
         now=utcnow_naive(),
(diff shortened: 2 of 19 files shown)


benkirk:
Added a second commit (`3ef8dd7`) at Ben's request: **Job Sizes hides a leading empty band.**

The 0 band can never fill for Nodes / CPUs — every job uses at least one of each — so it rendered as a permanent empty row plus a gap at the left of the chart. GPUs is the exception worth keeping: there the 0 band holds the CPU-only jobs (933K over 90 days on Derecho), so it survives on its own merit.

`_trim_leading_empty_bands()` sits in the shared histogram renderer, so Wait Times and Durations get the same treatment if their first band ever comes back empty. Two details worth a look:

- Emptiness is judged on `job_count` alone, never the displayed metric, so flipping Jobs / CPU-hours / GPU-hours can't shift the axis under the viewer. Zeros *inside* a distribution still render — that's what keeps the x-axis stable across filter changes.
- The trim runs before both the chart and the drill list, since `#jh-bar-<i>` and the table's `data-jh-bucket` are both positions in the same bucket vector. It returns a shallow copy — the envelope is a shared cache entry.

Full suite now **3360 passed, 30 skipped, 1 xfailed**. Browser-verified: Nodes starts at the `1` band with bar 0's drill carrying `min_nodes=1`; GPUs keeps its 0 band.

PR #394, opened 2026-07-28
Selector pills: Derecho leads, and white-is-active everywhere

benkirk:
Stacked on #393.

## Summary

Two defects in the strips that pick a machine or filesystem, plus one found while smoke-testing.

**1. Machine order — Derecho now leads.** `/user/jobs` and `/status/job-history` led with Casper. `job_history_machines()` sorted the engine keys, throwing away the order already expressed in config: `JOB_HISTORY_MACHINES` defaults to `derecho,casper` and `_warm()` inserts engines in exactly that order. It now returns insertion order, matching how the sibling helper `disk_scans.service.scan_capable_resources()` already preserves `FS_SCAN_RESOURCES` order — which is why Campaign_Store / Destor was already correct. The order is the deployment's, not the alphabet's.

**2. Pills follow white-is-active.** `.nav-pills` painted blue-is-active, the opposite of `.nav-tabs` and the `.btn-group` selector rows — so on `/user/jobs` the machine pills read backwards against the card tabs directly below them. Flipped `.nav-pills`, mirroring how `.nav-tabs` is written: custom properties for the active pill, explicit `:not(.active)` rules for the inactive one (Bootstrap gives no custom-property hook for that state).

The flip reaches all seven pill strips, so counter badges inside pills now invert with their pill — a stock `.bg-light` badge vanished on the now-white active pill. Both declarations need `!important`: the background to beat `.badge.bg-X`, the color to beat `.text-dark`, which Bootstrap ships as `!important`.

**3. Period pills.** The 30d/60d/90d/1yr group in `jobs_card.html` still used the `btn-primary`/`btn-outline-primary` pair — the same anti-pattern 19ab6c0 fixed for the metric pills, sitting in the same card and rendering blue-is-active right next to the machine pills this PR just fixed. Switched to `btn-outline-secondary` + `active`; the repaint in `nav-view-persistence.js` now toggles only `active`, since the base class no longer changes with state.

Left alone deliberately: `drp-toggle-custom` in `pickers.js` is a disclosure toggle for the custom-range panel, not a selector among peers.

## Test Results

Full suite: **3450 passed, 30 skipped, 1 xfailed** in 85s.

- `test_job_history_machines_sorted_when_enabled` → `test_job_history_machines_follows_config_order`, parametrized over both orderings so it pins insertion order rather than passing by accident on an alphabetical fixture.
- `test_card_fragment_marks_the_requested_pill_active` matches the new class pair and additionally asserts exactly one pill is active.

Playwright smoke against webdev, checking computed `background-color` / `color` on every affected strip:

| Surface | Result |
|---|---|
| `/user/jobs`, `/status/job-history` | **Derecho \| Casper**; active `#fff` bg + `#0057c2` text, inactive inverted |
| `/user/data`, `/status/filesystem-scans` | Campaign_Store / Destor order unchanged, colors flipped |
| Period pills | click still fans out across both machine subtabs and rewrites the "Open full view" `?days=` |
| `/allocations` pie tabs | now match the page tabs above them |
| Expirations sub-tabs | count badges legible on both blue and white pills |
| Allocation-tree resource types | badges invert correctly |

### DIFF
--- a/src/webapp/jobs/service.py
+++ b/src/webapp/jobs/service.py
@@ -49,16 +49,23 @@ def default_jobs_window_start() -> str:
 
 
 def job_history_machines() -> List[str]:
-    """Machines with a warmed job-history engine, sorted.
+    """Machines with a warmed job-history engine, in ``JOB_HISTORY_MACHINES`` order.
 
     Analogue of ``disk_scans.service.scan_capable_resources()`` — the
     single source for "which machines can the jobs UI offer?". Returns
     ``[]`` when the plugin is disabled, so nav items / tabs / route
     validation all degrade together.
+
+    The order is the deployment's, not the alphabet's: ``_warm()`` inserts
+    engines while iterating ``JOB_HISTORY_MACHINES`` (default
+    ``derecho,casper``), and dicts keep insertion order, so config order
+    survives — a machine whose engine failed to open just drops out. This
+    is what puts Derecho first in every subtab strip; sorting here used to
+    throw that away and lead with Casper.
     """
     if not is_enabled():
         return []
-    return sorted(get_engines().keys())
+    return list(get_engines().keys())
 
 
 def _resolve_queue_and_qos(
--- a/src/webapp/static/css/dashboard.css
+++ b/src/webapp/static/css/dashboard.css
@@ -1167,6 +1167,17 @@ h6, .h6 { font-size: 0.875rem; line-height: 1.35;     font-weight: var(--font-we
     margin-bottom: 0 !important;
 }
 
+/* Bootstrap's .card-header-tabs pulls the strip down over the header's cap
+   padding with a negative bottom margin — correct only when the tabs are the
+   header's LAST child. The usage card puts a custom date-range row after
+   them, and that negative margin dropped the tab bottoms ~16px into a ~30px
+   row, overlapping the range readout by half. Reset it whenever something
+   follows; the seamless-attach case above is unaffected because tab-content
+   is a sibling of the strip, not part of the header. */
+.card-header-tabs:not(:last-child) {
+    margin-bottom: 0;
+}
+
 .nav-tabs .nav-link {
     color: var(--ncar-blue);
     font-weight: var(--font-weight-semi-bold);
--- a/src/webapp/static/css/dashboard.css
+++ b/src/webapp/static/css/dashboard.css
@@ -1097,16 +1097,48 @@ h6, .h6 { font-size: 0.875rem; line-height: 1.35;     font-weight: var(--font-we
    Mirrors Unity's cascade: base color set first, :not(.active) overrides inactive.
    ========================================================================= */
 
-/* Pills — recolor to NCAR blue without changing the visual treatment.
-   Used for sub-navigation inside accordions and chart-selection rows where the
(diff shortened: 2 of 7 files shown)


benkirk:
Added a third UX fix in 50e49fc: **the custom date inputs now open under the button that summons them.**

On the resource-details usage card the Custom toggle is flush right (its pill group is `ms-auto`'d into the tab strip), but the inputs it revealed opened at the far left of the row below — the opposite corner from the button just clicked.

`drp_custom` gains an `align_end` flag defaulting to today's behaviour, because the macro's two callers genuinely differ:

- **Standalone card** (`resource_details_disk.html`) — `drp_pills` renders inline and Custom sits mid-row, so the inputs already open immediately to its right. Unchanged, verified pixel-identical.
- **Card header** (`resource_details.html`) — `drp_pills` is pushed right in the tab strip and `drp_custom` renders on its own row below. Passes `align_end=True`, so Apply lands directly beneath Custom.

`order-first` on the readout rather than reordering the markup, so both modes keep one DOM order and the readout stays the form's final field — only the visual order flips.

Note: the consequence of flushing the inputs right is that the current-range readout swaps to the **left** of that row (it was flush right). Both can't occupy the right edge, and putting Apply under Custom was the ask.

Before / after on the card header:

```
before:  [07/28/2024] → [07/28/2026] [Apply] .................. 2024-07-28 – 2026-07-28
after:   2024-07-28 – 2026-07-28 .................. [07/28/2024] → [07/28/2026] [Apply]
                                                                              ^ under Custom
```

Full suite still **3450 passed, 30 skipped, 1 xfailed**. Playwright confirmed Apply still submits — `start_date` edited to 2025-01-15 round-tripped into the URL with the `resource` / `scope` / `usage_tab` hidden fields intact.

benkirk:
Overlap fixed in 2484907, and verified down to mobile.

**Cause** wasn't the new layout — it was pre-existing. Bootstrap's `.card-header-tabs` carries a negative bottom margin (-15px here) to pull the strip down over the card header's cap padding, which is correct only when the tabs are the header's **last** child. The usage card puts the date-range row after them, so the tab bottoms ate ~16px of a ~30px row. Reordering the row just made it visible against the readout.

```css
.card-header-tabs:not(:last-child) { margin-bottom: 0; }
```

Same shape as the neighbouring `.nav-tabs:has(+ .tab-content)` rule. That case is unaffected — there the tab-content is a *sibling* of the strip, not part of the header. Class-plus-pseudo-class outranks Bootstrap's bare `.card-header-tabs`, so no `!important`. `resource_details.html` holds the only `.card-header-tabs` in the tree, so this is precisely scoped today while still generalizing.

**Measured `tab.bottom - row.top` in Playwright**, panel open and closed:

| Width | Overlap before | Overlap after | Layout |
|---|---|---|---|
| 1512 | **+16px** | +1px (the border line itself) | one row |
| 768 | — | −14px clearance | one row, readout left / Apply right |
| 390 | — | −14px clearance | wraps to 3 lines |

**Mobile needed no new rules.** The existing `<768px` net — `.main-content .d-flex:not(.flex-column) { flex-wrap: wrap }` — already wraps the form, so at 390px it degrades to readout / inputs / Apply stacked. Horizontal overflow measured **0** for the form, the input group, and the document at every width tested.

Full suite still **3450 passed, 30 skipped, 1 xfailed**.

PR #395, opened 2026-07-29
Jobs-tab activity timeline + charges as a first-class metric

benkirk:
> ⚠️ **Depends on [hpc-usage-queries#102](https://github.com/benkirk/hpc-usage-queries/pull/102)** (`jobs_timeseries` + QoS-weighted charges on every aggregate). That must merge and the image must rebuild before this deploys — SAM calls `JobQueries.jobs_timeseries`, which does not exist in an older plugin. Local containers are pinned to `b28a213`.

## Summary

The job-history card ships six tabs and five charts, and **all five are
distributional or categorical** — nothing on the card has a time axis, on a
page whose entire framing is a date window. This adds one, on the **Jobs** tab
(the only tab that had no chart), driven by the live job filter set.

SAM already renders this chart for *summaries* (the stacked-by-user Usage
Trend on resource-details), but that reads `comp_charge_summary` and cannot
answer anything about queues, job sizes, wait times or exit status. This one
honours the whole `jobs_search` filter set, so the chart and the table beneath
it can never disagree — the plugin pins `total_count == jobs_count`.

## What's here

**Activity timeline** — a stacked bar chart above the Jobs table: one bar per
calendar band, segmented by the window's top-10 owners over an aggregated
"Others" base, with pills for metric (jobs · CPU-hours · GPU-hours · charges),
granularity (day · week · month) and owner axis (user · project). Bars drill
into the period; legend entries open the entity's quick-view modal.

**`charges` on all six panels**, not just the new one, so there is one metric
vocabulary and the shared `metric:jobs` persist family stays valid. Two things
fell out of that worth review attention:

- `charts._JOBS_METRIC_KEYS` now maps to key **tuples** — `charges` is the one
  metric backed by two plugin keys (`cpu_charges` + `gpu_charges`). A single
  `_jobs_metric_value()` accessor means a multi-key metric can't be read as a
  single key somewhere and silently render as zero.
- The metric pills became a shared `metric_pills()` macro rather than a fourth
  hardcoded copy, and `JOBS_METRIC_NOUNS` replaces an inline label dict that
  would have `KeyError`'d on the new metric.

## Design notes

**Granularity is server-chosen, with override pills.** The explorer permits an
unbounded window (clearing both date fields — ~1,500+ days on derecho), so a
fixed `day` default would blow the bar budget or trip the plugin's 400-band
`ValueError`. `_auto_period` picks the finest granularity under 120 bars; an
explicit `?period=` is honoured **only while it fits**, so a stale `day` from
localStorage can't be replayed against a five-year window. Over-budget pills
render **disabled with the bar count as the reason** rather than vanishing —
a pill that disappears reads as a bug.

This is load-bearing, not polish: 7 → 180 bands measured **+54%** on a fixed
4.28M-row window (plugin PR #102).

**Collapsed on cards, open on the explorer.** Jobs is the *default* tab on four
surfaces, so an unconditional aggregation there would be paid by every visit
that only wanted the table — a worse deal than the five chart tabs, which are
opt-in because you clicked them. The fetch binds to `shown.bs.collapse` **on
the div**, never a click: `nav-view-persistence` re-opens a saved collapse
without firing click, and a click trigger would leave a restored-open panel
permanently empty. (Verified in the browser — see below.)

**Band drills replace the window, not narrow it.** Unlike a histogram band
(`min_param`/`max_param`), a time band replays through `start`/`end` — the
window filters *are* this dimension — so `_band_drill_url` overrides the pane's
dates rather than appending to them.

**Charges are not proportional to hours.** `qos_factor` is a genuine `0.0` for
the `uncharged` QoS, so a charges view legitimately shows an empty bar where an
hours view shows work. The panel carries a caption saying so; without it an
empty bar reads as missing data.

## Two defects the test suite could not catch

Both were found only by clicking, and are fixed in `8f170a3`:

1. **The legend was a silent no-op.** It used `#job-user-` row sentinels,
   assuming `activateOwningTab()` would hop panes. It doesn't — `ROW_SENTINELS`
   are resolved by `openEntityRow()` *scoped to the clicked chart's tab-pane*,
   and this chart is in the Jobs pane while those rows are in the lazily-loaded
   By User / By Project panes. Now points at the entity **modal route**, the way
   the status dashboard's stacked-area chart already solves it; works from any
   pane, needs nothing pre-rendered. `link_entities` gates on
   VIEW_USERS / VIEW_PROJECTS, matching the tables.
2. **The timeline eager-loaded into a hidden pane.** The explorer round-trips
   `active_tab` server-side, so an Apply from Wait Times re-rendered with Jobs
   hidden and *still* fetched — two aggregate scans for a panel nobody could
   see. Now inherits the Jobs pane's trigger via `_trig('jobs')`.

**Known limitation, pre-existing and deliberately not fixed here:** when the
client restores a different tab from `tab:jobsExploreTabs` localStorage, the
server can't know, so the eager load still happens. The Jobs *table* has always
behaved this way — it's the URL-vs-localStorage race already documented for
these lazy panes, and it belongs with the tab-state work.

## Test Results

**3,483 passed, 30 skipped, 1 xfailed** (+33).

- `test_webapp_jobs_charts.py` (+12): all four metrics render; charges sums
  BOTH plugin keys; empty/all-zero placeholders; `#jt-bar-` sentinels on
  populated bands only; legend identical across bands; "Others" is the
  derivable remainder and absent when owners cover the total; legend opens the
  modal (not a dead row sentinel) and is unlinked without permission;
  period / link-flag / entity-kind / count-vector all join the cache key.
- `test_webapp_jobs.py` (+21): the granularity budget table, explicit-choice
  clamping, unknown-value leniency (incl. `quarter`/`year` — PeriodGrouper's
  vocabulary, deliberately not ours), disabled-not-hidden pills, span parsing
  incl. the open-ended case, band-drill override, and `charges` being rankable
  everywhere it is offered.
- **Route-map snapshot regenerated: +3 routes, nothing else.**

### Browser smoke (Playwright, Quick Login as `benkirk`)

- Status → Job History: both cards collapsed and **not fetched** on load — the
  table-only path costs nothing extra.
- Expand → chart, 4 metric pills, 3 granularity pills (Day 87 bars / Week 13 /
  Month 3), owner pills; 482 bar links, 20 legend links, 87 period rows.
- **Reloading with the collapse saved open re-fired the fetch** — confirming
  `shown.bs.collapse` is the right binding.
- Legend click → user modal opens, populated. Project axis → 20 project-modal
  links. Bar click → correct period row expands with a 14-row owner tier; owner
  drill lazy-loads 101 jobs scoped to `2026-05-05 … 2026-05-05`.
- Charges caption renders; Week → 13 ISO bands (`2026-W18…`) with the metric
  preserved; Charges persisted across pages, proving the shared family.

### Cross-check against real data

`jobs_timeseries`, `jobs_usage_by` and `jobs_histogram` agree **exactly** on
derecho over 14 days — 173,031 jobs / 75,037,615 cpu-hours / 72,892,958
cpu-charges — so the large per-day totals in the dev dataset are the data, not
a join fan-out.

## Deploy notes

- Merge + rebuild hpc-usage-queries#102 first.
- No new cache bucket; `period` joins the existing `jobs` / `jobs_recent` cache
  opts. Fan-out grows (~8 more keys per filter combination), so re-check Redis
  `INFO memory` / `evicted_keys` before the main promotion, per
  JOBS_EXPLORER_CHARTS.md C7.


benkirk:
Findings from reviewing the plugin half ([hpc-usage-queries#102](https://github.com/benkirk/hpc-usage-queries/pull/102)). Posting rather than pushing, so this PR stays one person's work.

**None of these are merge blockers, and none are compatibility breaks.** The plugin changes require zero consumer updates — this branch reads `total_count`, `null_count`, `bands[]`, `start`, `end`, `owners_by`, all unchanged. The pre-existing ordering constraint stands: #102 must merge and the plugin container rebuild before this goes live, since `routes.py:522` sends `sort_by='charges'`.

---

### C1 — the usage panel offers a `charges` metric pill but has no charges column

`_METRICS` (`routes.py:490`) is one vocabulary across every panel, and the By User / By Project panel parses from it (`routes.py:914`, via `_parse_metric(_DEFAULT_METRIC_PIE)`). So `?metric=charges` is reachable there. Today:

- the **pie** ranks and sizes by charges correctly — `_jobs_metric_value` maps `'charges' → ('cpu_charges','gpu_charges')` (`charts.py:1276`), and its "Other" remainder is metric-aware (`charts.py:1660`);
- the **table** beside it has four fixed columns — Entity / Jobs / CPU-hours / GPU-hours (`jobs_usage_panel.html:79-82`, rows at `:122-124`). No charges column, so you rank by a number the table can't show;
- the totals strip omits charges (`:53-55`);
- no column carries the `sort-desc` initial-sort indicator when `metric == 'charges'` — the `{% if metric == ... %}` conditions only cover jobs / cpu_hours / gpu_hours;
- the "Other" row is labelled `…(charges)` via `JOBS_METRIC_NOUNS[metric]` (`:154`) while its three numbers are jobs / CPU-h / GPU-h.

`_usage_other` (`routes.py:831`) hardcoding `('job_count', 'cpu_hours', 'gpu_hours')` is one symptom of this, not the whole thing — fixing just that tuple would leave the missing column.

Docstrings to update alongside: `jobs_usage_panel.html:19` (the `other` key list) and `:21` (metric vocabulary, still `'jobs' | 'cpu_hours' | 'gpu_hours'`).

### C2 — `routes.py:495-497` cites a benchmark that has been retracted

The comment justifying `_MAX_TIMELINE_BARS = 120` reads:

> band width is a real cost knob — 7 -> 180 bands measured +54% on a fixed 4.28M-row window (plugin PR #102)

That +54% was a measurement artifact: the three periods were timed **sequentially**, so buffer-cache warming rode along with band count. Re-measured with the periods interleaved inside one loop, on PG 18 / casper_jobs (21.0M jobs):

| window | period | bands | min | vs month |
|---|---|---|---|---|
| 180 d (4.31 M rows) | month | 7 | 6735 ms | 1.00× |
| | week | 26 | 6848 ms | 1.02× |
| | day | 180 | 7390 ms | **1.10×** |
| 730 d (16.6 M rows) | month | 25 | 8909 ms | 1.00× |
| | week | 105 | 9505 ms | 1.07× |
| | day | 730 | 14699 ms | **1.65×** |

So ~10% at 180 bands, not 54%. The `O(bands/2)` concern is real, it just doesn't bite until several hundred arms.

**Keep 120.** The real justification is the one the same comment already gives — *"180 bars is already past what an 18in axis can show"*. That's a legibility limit and it stands on its own; it just isn't a cost limit.

The adjacent line, *"The plugin hard-caps its CASE ladder at 400 bands and raises"*, also needs updating: the cap is now path-dependent — 400 on the `jobs`-scan path, 1200 on the new `daily_summary` fast path, which has no ladder at all.

### C3 — `service.jobs_timeseries`'s cost docstring is now wrong

It says the timeline *"is the most expensive panel on the card — which is why its host keeps it behind a collapse."* With the fast path in #102, a scoped 180 d daily series is ~15 ms end to end against ~7.4 s scanning `jobs` (~470×). It routes to the rollup whenever the filter set is expressible in `(date, user_id, account_id, queue_id)` — which is exactly the dashboard's scope — and falls back for `qos`, `exit_status`, `job_id`, `name` and any `min_*`/`max_*` bound.

Whether to un-collapse the panel is a UX call, not a cost one any more.

### C4 — test fixtures predate the charges metric

`_sample_hist()` / `_sample_usage()` (`tests/unit/test_webapp_jobs.py:1608`) and `_hist()` / `_usage()` (`tests/unit/test_webapp_jobs_charts.py:27`) build envelopes with no `cpu_charges` / `gpu_charges` keys. So every charges-metric assertion currently exercises the `.get(k) or 0` fallback rather than real values — a swapped or dropped charge key would pass unnoticed. `_ts()` (`test_webapp_jobs_charts.py:378`) already carries them, so it's just those four.

Worth adding real values there and then asserting on the C1 column and the C1 "Other" remainder.

---

For context, the plugin side also picked up two fixes worth knowing about, both of which affect what you'll see rather than what you call:

- `sync/summary.py` inner-joined `job_charges`, silently dropping chargeless jobs from `daily_summary` and hence from `daily_summary_report`. Now a LEFT join. (0 of 21.0M rows affected on casper — `trg_ensure_job_charge` makes it impossible in production — but the rollup shouldn't depend on a trigger for its arithmetic.)
- `jobs_timeseries` now re-binds its statements to the resolved window, so a band's `jobs_search` replay can't be broken by a job synced mid-query.

benkirk:
All four addressed in `ba07cc1`. Confirmed every claim against the code first; all four held, and two were worse than described.

### C1 — fixed

Confirmed end to end. `_parse_metric` (`routes.py:682`) takes no panel argument and `metric_pills()` renders all four for every caller, so `?metric=charges` is reachable, and the server really does re-rank and take the top-N cut in charges. The pie was already right.

Added the Charges column (header + sort guard, cell, Other row, totals strip, `colspan` 4→5) and gave `_usage_other` the full vector via a new `_USAGE_METRIC_KEYS`. One deliberate asymmetry: the remainder is *computed* in all five keys but its **visibility** still keys on count/hours only. An all-`uncharged`-QoS tail is a real tail, and no row can carry charges without hours, so folding charges into the `> 1e-9` test would only add float noise.

**One thing the writeup missed:** `routes.py:489` asserts *"templates carry a caption saying so"*, but that caption existed **only** in `jobs_timeline.html`. It was absent from both `jobs_usage_panel.html` and `jobs_histogram.html` — i.e. the two panels where an empty charges view is *least* self-explanatory. Added to both.

### C2 — corrected, cap kept at 120

Agreed, and thanks for catching that the committed bench script measured sequentially too — that is why it never reproduced its own table.

Rewrote `routes.py:494-497` to rest the cap on legibility, which is the justification that survives, and flagged the retraction explicitly rather than silently swapping numbers. Also corrected the 400-band claim to be path-dependent (400 scan / 1200 summary). Second occurrence fixed at `docs/plans/JOBS_TIMELINE.md:74`, where `+54 %` was the sole support for *"auto-coarsening is load-bearing"* — that conclusion now rests on the axis limit.

### C3 — reworded, collapse kept

Rewrote both this docstring and `_render_timeline`'s, which repeated the same two claims. Framed as conditional rather than swapping one absolute for another: cheap on a card's scope, expensive under exactly the explorer filters that force the scan.

Also dropped *"its host keeps it behind a collapse"* — `routes.py:1507` sets `timeline_open=True` on the explorer deliberately, so that was already false for one host.

**Keeping the collapse.** The panel is only cheap on the subset of views where it is least interesting, and see P2 below.

### C4 — fixed, and it was worse than stated

Beyond the four fixtures, there was **no charges render coverage at all**: `test_webapp_jobs_charts.py:62` looped all metrics for the histogram but omitted `'charges'`, no pie test used it, and no route test ever fetched `metric=charges`. `test_charges_is_a_first_class_metric_everywhere` asserted two module-level dicts despite its name. So the weak fixtures were not masking failures — nothing exercised those paths.

Fixtures now carry charges deliberately **not** proportional to hours, with an uncharged row/bucket in each, so a metric that read `cpu_hours` where it meant `cpu_charges` renders differently instead of identically. Added the histogram charges case, a pie case asserting the remainder stays in charge units, and route-level `metric=charges` fetches asserting the C1 column and remainder.

Extended the "first-class" test to require every metric be **renderable** (`_METRICS ⊆ _JOBS_METRIC_KEYS`) and its keys computable in the remainder. That is the membership gap C1 slipped through: `charges` was requestable and rankable, so both existing assertions passed, while nothing could display it.

---

3,492 passed / 30 skipped / 1 xfailed. Route-map parity untouched. Browser-smoked against a container built on the refined plugin (fast path present, `_MAX_SUMMARY_BANDS=1200`): Charges sorts on the raw float, Other reports 285M charges, drill-down spans `colspan=5`, caption on all three panels and only on charges, timeline unchanged at 743 band sentinels, zero console errors.

Two findings going back to #102 — posting there.

benkirk:
Re-verified against the **merged** plugin (`965ebac` on `main`), not the branch build. No ref bump was needed here — every `HPC_USAGE_QUERIES_REF` pin defaults to `main`, so a plain rebuild picks it up.

Warm timings on the rebuilt containers, `/status/job-history`:

| fragment | branch build | merged build |
|---|---|---|
| timeline | 62.4 ms | **60.9 ms** |
| jobs table (beside it) | 752.0 ms | **761.7 ms** |
| by-user | — | 53.3 ms |

Identical within noise, which is what a squash of the same commits should give. Cold-cache after the rebuild was 769 ms / 2311 ms — that spread is matplotlib and the aggregation cache warming, not the query.

Also re-checked in the browser: timeline open by default with 692 band sentinels, the second machine card still lazy behind its tab, the Charges column present and carrying the initial-sort indicator, the "Other (beyond top 25 by charges)" row reporting 298M in charge units against a 638M total, and the uncharged-QoS caption rendering. Zero console errors.

Worth noting for anyone reading the earlier red checks on this branch: CI builds the plugin from `main`, so every pre-merge run failed by construction — `JobQueries.jobs_timeseries` did not exist there yet. The current run is the first meaningful one, and it is green across all nine checks.

PR #408, opened 2026-07-31
Front-end regression net: shell contracts + Playwright console sweep

benkirk:
Supersedes #380, which was closed: its plan predated #406 (the Node 20 Actions-runtime migration), so its entire CI section targeted workflows that no longer exist. Review also found two substantive errors in that plan — the sweep's route list invented `/status/reservations` (never a route) while omitting seven real pages, and the rate limiter's global 200/min default would have 429'd the sweep. Both are fixed here.

## What this adds

Two complementary layers plus the CI workflow that runs the browser half. `docs/plans/FRONTEND_TEST_NET.md` is the decision record, including why a JS unit harness (vitest/jsdom) was rejected.

| Layer | Catches |
|---|---|
| `tests/unit/test_modal_shell_contract.py` | dangling `data-bs-target` — **silent in the browser** |
| `e2e/` (Playwright) | dangling `hx-target`, JS exceptions, script-order breaks |

Neither subsumes the other, and that was verified rather than assumed: htmx *does* `console.error("htmx:targetError")` on a dangling `hx-target`, but a dangling `data-bs-target` produces no console output at all — `htmx-config.js` listens for `responseError`/`sendError`, never `targetError`. A browser-only net would miss Bootstrap-only affordances entirely.

## Part 1 — shell contracts

The template tree has **no dynamic `{% include %}`**, so the include graph is fully static and this can be a real assertion rather than only a ratchet:

- **Static** — each of the 37 page-template closures resolves every `#id` it targets. Zero violations today. Macro-generated shells count: 43 `modal_scaffold()` call sites never emit a literal `id="..."`, which an earlier naive pass missed.
- **Ratchet** — 25 htmx-loaded fragments reach for shells their *host page* must supply. Invisible to static analysis, so the set is pinned by equality.
- **Rendered** — walks the exact chain PR #378 broke through the Flask test client.
- **Script order** — `actions.js` defines `window.registerAction`/`revealCard`; five later scripts call them at eval time. Undeclared and load-bearing.

## Part 2 — `e2e/`, outside `tests/`

`tests/conftest.py`'s guard hard-exits unless `SAM_TEST_DB_URL` points at the allowlisted mysql-test container. Conftest files load only from a test file's ancestor directories, so keeping `e2e/` a *sibling* means the guard is never loaded and never has to be weakened. Nothing under `e2e/` imports sam/webapp/system_status — that boundary is what lets CI install a bare Python instead of building the conda env.

The 22 swept routes are **derived** from `tests/unit/snapshots/dashboard_route_map.json`, so a new top-level tab enters the sweep automatically instead of being forgotten. The console allowlist is **empty**: every dashboard page currently loads with zero console output, and a ratchet stops any future entry from rotting.

## Part 3 — `browser-smoke.yaml`

Its own workflow, not a step bolted onto `test-install.yaml` or `sam-ci-conda_make.yaml` — both exist to prove an *install path* works, and both say in their own comments that the suite is deliberately not run there. A front-end regression turning either red would point reviewers at the installer. Modeled structurally on `sam-ci-conda_make.yaml` but with `setup-python@v6` instead of conda (seconds, not minutes).

Post-#406 conventions throughout: `checkout@v7`, `setup-docker-action@v5`, the `[skip ci]` guard, `concurrency`, `paths-ignore`, `workflow_dispatch`, and no `_` prefix (that now means "reusable workflow"). Every action pin is one #406 already uses, so dependabot's weekly grouped PR maintains them.

**Rate limiting.** `webapp/limiter/__init__.py:73` installs a *global* `default_limits` of 200/min keyed to `user:<username>`, Redis-backed and shared across gunicorn workers. The sweep is one actor across 22 fragment-heavy dashboards; a 429 arrives as an htmx error indistinguishable from a real regression. The workflow raises `RATELIMIT_AUTHED` in `.env` rather than disabling the limiter, so its code path stays under test.

**Pinning.** `pytest-playwright` goes in a new `[e2e]` extra with an upper bound. `.github/dependabot.yml` is scoped to github-actions on purpose ("Python dependencies … are a separate decision"), so nothing bumps this for us. Pinning the driver also pins the Chromium build `playwright install` fetches.

## Test Results

- **Part 1** — 61 passed, 2 skipped. The skips are `/user/data` and `/user/jobs`, which `abort(404)` when their optional plugin is absent: the designed response, now encoded rather than papered over.
- **Part 2** — **34 passed, 0 skipped, 52s** against the compose stack on :7050. Getting to zero skips took three rounds of selector work; the pencils live in collapsed accordions and the project-details openers are `hx-get` anchors, not `data-bs-target`, so the first drafts were silently skipping the guards that matter.
- **Both nets proven to bite.** Removing the `allocation_modals.html` include that PR #378 added fails the Python tier across 21 pages naming the missing shell, and fails the browser tier with `/status/derecho does not ship #editAllocationModal` plus a timeout waiting for a modal that can never open — the original silent no-op, reproduced and caught.
- `actionlint` reports **zero** findings on the new workflow.
- Bare `pytest` unchanged: 3,988 collected, none from `e2e/`.

## Note for the reviewer

`docs/TESTING.md:5` and `CLAUDE.md` both say "~3,100 collected tests"; the actual count is now 3,988. Left alone as out of scope for this PR.


benkirk:
reopening to re-fire the pull_request event (no Actions check suite was created on open)

PR #424, opened 2026-08-05
XRAS reimplementation: the read API, action capture, all six handlers, and the operator surface

benkirk:
## Summary

Reimplements NCAR SAM's XRAS integration in Python/Flask. It moves the read API off
the legacy Java stack byte-for-byte, stands up the write endpoint, ports **all six
action handlers**, and gives operators a first-class surface for the result.

That is **Phases 1–4** of `docs/plans/XRAS_REIMPLEMENTATION.md`, shipped as one
ordered commit series across five sprints. The measured production data, the wire
contract and the deliberate divergences live in that plan; the as-built records are
[`XRAS_SPRINT_A.md`](docs/plans/XRAS_SPRINT_A.md),
[`XRAS_SPRINT_B.md`](docs/plans/XRAS_SPRINT_B.md),
[`XRAS_SPRINT_C.md`](docs/plans/XRAS_SPRINT_C.md),
[`XRAS_HANDLER_REFACTOR.md`](docs/plans/XRAS_HANDLER_REFACTOR.md) and
[`XRAS_STRESS_AND_SCHEMA.md`](docs/plans/XRAS_STRESS_AND_SCHEMA.md).

⚠️ **Nothing dispatches yet.** `XRAS_ACTIONS_CAPTURE_ONLY` defaults **on** and flips
only at cutover, when XRAS repoints its base URL. The handlers are built, tested and
dormant. See *The safety interlock* and *What is left* below.

## What landed

| Area | What |
|---|---|
| **Read API** — Phase 1 | The six GET endpoints under `/api/xras/v1`, byte-compatible with legacy, covering **94% of production traffic**. Ported against base tables rather than the `xras_*` views, which are unusable as-is (§ 4.2). `login_or_token_required` gains `roles=` / `deny=`, closing an authz gap where `g.api_key_roles` was populated but never enforced. Latency is 20–200× better than legacy on every endpoint |
| **Action capture** — Phase 2 | `xras_action_log`, `XrasActionSchema` (seven nested schemas written against real payloads), and `POST /api/xras/v1/actions`, which authenticates, parses, audits and answers. Both URL forms are mapped: the bare one every real post uses, and the segmented one the ACCESS spec documents |
| **The handlers** — Phase 3 | All six services, dispatching on the **pair** `(actionType, does the project exist)` as legacy does: Extension, Supplement, Adjustment, New, Update, and Transfer — which is a *registered* handler that parks with a reason rather than an absent one that looks like an oversight. Every handler follows assemble → check once → execute, so a rejection happens before any transaction opens and "nothing was written" is true by construction |
| **The error vocabulary** | All **31** legacy error strings reproduced byte for byte, typos included, because XRAS administrators read them directly and have learned them — plus **3** new ones (`ambiguous_role`, `ambiguous_contract`, `adjustment_would_go_negative`) for conditions legacy crashes on rather than reports, each marked as non-legacy in the module. Accumulated in an insertion-ordered, **deduplicating** container mirroring Java's `LinkedHashSet`, so three resources missing an amount produce one line rather than three |
| **Operator surface** — Phase 4 | A 4th top-level page on the Allocations dashboard: filterable action log with self-excluding facet chips, a detail modal, and one-click **replay**. Plus `VIEW_XRAS` / `MANAGE_XRAS`, split on *what the data is* rather than read-vs-write |
| **Activation worklist** | XRAS projects arrive `active = 0` and a human activates them. `xras_activation_event` is an append-only event table whose state is **derived, never stored**, so a dismissal re-opens when new information arrives and nobody is notified twice |
| **`sam-admin xras`** | list / show / payload / summary / replay / `--validate-mapping`, with the same filters and the same query layer as the web page, so the two cannot drift |

Supporting work: a payload scrubber and a dev seeder (`scripts/xras/`), `--api xras`
in the parity harness (its first byte-exact comparator), a `purge_xras_action_log`
rule in the snapshot anonymizer, and a `Makefile` fix — `make docker-build` never
rebuilt `mysql-test`, so an amended init script appeared everywhere *except* where
pytest looks.

## The bug worth reading this PR for

`resolve_resource` read `resources[].key`. **XRAS has never sent that field.** All six
resource-bearing corpus fixtures send `resourceRepositoryKey`, this repo's own wire
documentation names it correctly (§ 3.1), the schema declares it under that name, and
unknown keys are dropped on load. So through the real pipeline the key was always
`None` and every resource on every Supplement, Adjustment, New and Update reported

```
No resource found in SAM corresponding to key
```

with nothing after it — roughly **36% of production traffic** failing on day one of an
abrupt cutover, with an error naming no key because there was none to name.

**Why a whole sprint of tests missed it.** Every test built its own `resources[]`
entries as `{'key': ...}` — five handler modules, the error-coverage matrix, and the
oracle's re-targeting helper, whose docstring reads *"Shape untouched"* while replacing
the one field that mattered. Every layer that could have caught it substituted the
wrong shape first.

**The fix is the check, not the field name.** `tests/unit/test_xras_wire_vocabulary.py`
asserts that every wire field name the handlers read is declared by an XRAS schema — an
AST walk over `get_field(...)` literals against the union of all seven schemas. It named
the bug on first run as `{'key': ['_fields.py:159']}`.

It was found by driving a real payload through the real route, which is exactly what the
stress tier exists to do.

## Behaviour that differs from legacy

The headline improvement, and the one thing needing sign-off before cutover. Legacy
answers 500 with an opaque timestamp for both a malformed body and a failed validation,
and a bare 200 for an action it silently parked — so its 200/500 split cannot
distinguish "processed" from "quietly deferred".

| Condition | Legacy | Here |
|---|---|---|
| Malformed JSON | 500, opaque | **400** + the reason, row with `action_type` NULL |
| Validation errors | 500, opaque | **422** + the accumulated, ordered error list |
| Oversized body | 500 | **422** saying so, row kept with the payload marked |
| Processed | 200 | 200, row `processed` + `service` |
| No serviceable handler | 200, no trace | 200, row **`manual`** + `service` + `outcome_reason` |

⚠️ **Broker retry behaviour on 4xx is unknown.** Confirming this with
`allocations@access-ci.org` is a prerequisite of the POST cutover and has external lead
time.

Reading the Java turned up a set of legacy defects. Each was decided individually and
recorded — the full table is `XRAS_SPRINT_C.md` § *Deviations*:

- **Fixed**, because they are plainly wrong: duplicate current PI roles resolved by array
  order, so a coin flip decided who leads a project (defect 1); `isServiceable` testing
  `"Adjust"` against a wire that sends `"Adjustment"`, which has kept
  `AdjustProjectActionService` dark for its entire existence (defect 4); the copy-pasted
  `> 0` gate that dropped the negatives an adjustment exists for; and an Update whose
  guard meant it never actually moved the lead or admin.
- **Not ported, and warned instead**: the `UNDO AUTO/DEFAULT` compensating adjustment,
  which is dead code — writers use `.name()`, the detector compares `.getValue()`, and
  production holds zero `UNDO` rows of either spelling (defect 5); and an Update that
  silently re-activates a project a human has not yet approved.
- **Surfaced rather than repaired**: a person assigned a role but excluded from the
  project's own roster (defect 3). It is warned about because that warning is the only
  evidence anyone has that the situation occurs.
- **Turned into a reviewable 422 rather than copied or silently fixed**: the live data
  conditions legacy 500s on. Three contract cores genuinely collide (`1049089`,
  `1744587`, `2146709` — re-measured after a fresh clone, still exactly those), where
  legacy's `LIKE '%core'` + `uniqueResult()` guarantees a 500 and we report
  `ambiguous_contract` naming the candidates. Organization 158 matching two mnemonic
  codes (defect 2) lands the same way.

## The safety interlock

`XRAS_ACTIONS_CAPTURE_ONLY` defaults **on**. While it is on, `POST /actions` records the
row and returns 200 without dispatching, and querying `status='received'` gives operators
the capture backlog.

`XRAS_ACTIONS_ENABLED` is a *separate* per-type lever, and not a rollout mechanism: all
six handlers ship enabled in one deploy, and it exists so a misbehaving payload class can
be parked by config at 3am instead of by revert. A disabled type takes the audited manual
path rather than being dropped.

**Replay never dispatches, unconditionally.** Sprint B coupled it to the capture flag;
Sprint C reversed that. XRAS owns the retry, so a replay that applied would race a resend
with no idempotency key between them. Replay re-parses and re-validates, which makes the
button a standing regression check of the schema against the harvested corpus.

## Audit-row schema

`xras_action_log` gained three columns, each with written evidence in
[`XRAS_STRESS_AND_SCHEMA.md`](docs/plans/XRAS_STRESS_AND_SCHEMA.md) § *Verdicts*:

| column | why |
|---|---|
| `action_id` | Three identical posts produced three rows identical in every filterable column; `actionId` lived only as bytes inside `raw_payload`. The cost is asymmetric and measured — Extension is 60% of traffic and a double post writes nothing, Supplement is 15% and adds a full increment |
| `service` | Four parking causes produced **byte-identical** rows. A test asserted that equality; it now asserts the difference |
| `outcome_reason` | Transfer's "why" is 200 characters written for whoever reads the row at 3am, and it died in an ephemeral k8s log |

A fourth candidate (`warnings`) was **declined** with reasons. Two more gaps were closed
in code rather than schema: `raw_payload` and `error_messages` were unbounded into `TEXT`,
and under `STRICT_TRANS_TABLES` an oversized value does not truncate — it raises, and the
audit row is lost. Reproduced live before the guard existed.

## Test results

**5,309 collected, 5,270 in the default run**, ~90 s under xdist (`staging` baseline
4,404). Two tiers are gated off and run on request: **22 perf**, **17 stress**.
`docs/TESTING.md` carries the tier breakdown and is the number to trust.

About 900 tests are XRAS-specific across 39 files. Three of them are structural gates
rather than behaviour tests, and each exists because the failure it prevents is *silent*:

- **`test_xras_wire_vocabulary.py`** — field names read must be names the schema declares.
- **`test_xras_transaction_seam.py`** — `management_transaction` must have exactly one
  import site under `sam.xras`. A runtime `vars(module)` scan, deliberately not a grep
  (blind to re-exports) or an AST walk (blind to runtime rebinds), plus a `session.commit`
  spy that catches a commit reached by *any* path. Verified to fail by reinstating the
  import.
- **`test_xras_error_coverage.py`** — all 34 error strings are exercised or declared
  unreachable with a reason.

The leak check (`allocation_transaction` / `xras_action_log` / `project` rows created
today) returns **0** after every run, including under `-n auto`.

## Notes for review

- **The two tables still need a DBA to run the DDL in production.** The prod writer holds
  no DDL grant and Alembic manages only `system_status`. Dev and CI get them from tracked,
  self-retiring `initdb.d/zz-90-*.sql` and `zz-91-*.sql`; **staging needs both run by hand
  once** (`init-rds.sh` restores the raw `.xz` with no initdb hook). One ticket carries
  both — a second costs another round of lead time. The ticket is now a **transcription of
  `zz-90`**, not a design question.
- Picking the tables up locally needs
  `docker compose --profile test down -v && make docker-build && make docker-up` —
  `make docker-down` has no `-v`. Verified end to end on a clean rebuild + clone.
- The audit row is committed **outside** `management_transaction`, which rolls the whole
  session back on exception. That is also why the API and stress tests clean up
  explicitly: a per-test SAVEPOINT cannot undo a write on another connection.
- Route tests **cannot use factories** — the route reads `db.session` on its own
  connection and sees only committed rows, so a factory project silently turns a scenario
  into "no service matched".
- `anonymize_sam_db.py` **purges** `xras_action_log` rather than scrubbing it —
  `raw_payload` is a verbatim POST body full of PII and the obfuscated dump is a committed
  public LFS blob.
- The legacy-compat GET blueprints remain additive-only; response bytes are a contract.
- ⚠️ **`API_KEYS_SAMUEL` must never be set in config.** `_verify_api_key` checks config
  first and never falls through to the DB on a hit, and config identities carry no roles —
  so setting it would make every XRAS request 403 while legacy kept serving happily.
  Pinned by a test.

## What is left

**One code item**, found by auditing this branch against the deployed `ROOT.war` and
58 days of legacy access logs (full write-up in the review comment below):

0. **`POST /api/xras/v1/roles/{requestNumber}/{role}/{username}` is unported.** It is
   endpoint **#7** in `XRAS_REIMPLEMENTATION.md` § 2.1 — `RoleServiceController.setUserRole`,
   a **write** that reassigns the project lead — and it is mapped by `web.xml` today. Zero
   traffic in 58 days, but cutover repoints one base URL, so it would 404 where legacy
   200s, silently. The decompiled contract is in the comment; note the failure statuses
   are **403**, not 400, and success is a **truly empty 200** with no envelope. Porting it
   reuses `Project.update(project_lead_user_id=…)` — no new write primitive.
   Landing alongside it: a WARN log for any unmapped path under `/api/xras/`, so the next
   one of these is a grep rather than a silence.

Then four gates, in order:

1. **The DBA ticket** for both tables — the only item with a third party's schedule.
2. **`sam-admin xras --validate-mapping` clean**, *then* `--api xras` against the deployed
   host. Currently **not** clean and filed as a data decision: 13 mapping rows, **11 active
   resources XRAS cannot name** (Gust, Gust GPU, GLADE user/work, …), 6 mappings pointing at
   decommissioned resources. ⚠️ Adding a mapping **changes GET response bytes**, so this
   closes *before* the parity run, not after.
3. **The 400/422 contract confirmed** with `allocations@access-ci.org`.
4. **XRAS repoints its base URL** — the cutover, and the moment
   `XRAS_ACTIONS_CAPTURE_ONLY` flips to `0`.

Deliberately out of scope:

- **SMTP.** Legacy keeps mailing until cutover step 4, and the pending-activation card is
  the accepted substitute trigger — so Notify records the event and shows an explicit
  "email delivery is not implemented" dialog rather than a success toast implying mail moved.
- **`compose.yaml` sets no `TZ`** while `helm/values.yaml` sets `America/Denver`, so local
  dev and CI run UTC against ~123 `datetime.now()` call sites. Its own change, its own
  full-suite run.
- **The wire shapes the corpus has never carried.** `Co-PI` vs `CoPi` is now **closed** —
  tested across three spellings, and membership ignores `roleType` entirely, so the
  spelling cannot matter. `Renewal` and `Advance` are exercised synthetically. `Transfer`
  parks by design. What remains is production capture after cutover.

⚠️ **Cutover is abrupt.** XRAS repoints one base URL and all six handlers go live at
once — there is no observation window and no per-type rollout. That constraint is why
this PR carries a stress tier, three audit columns and three structural gates rather than
just handlers.


benkirk:
## Endpoint-coverage audit against the legacy Tomcat host

Checked this PR's central claim — that it maps the XRAS→SAM surface — two ways:
against the legacy access logs, and against the **deployed `ROOT.war` itself**,
which is the authoritative servlet mapping rather than anyone's notes.

**Verdict: every endpoint XRAS actually calls is ported. One mapped legacy
endpoint is not — and it is a write.**

### Evidence base

The live logs only reach 2026-07-09, but `/tomcat/.snapshot/weekly.2026-07-12_0015/`
carries `access.log-20260612.gz` onward, so the usable window is
**2026-06-12 → 2026-08-08 — 58 days**, roughly double what § 1.1 measured.

| Source | What it gave |
|---|---|
| `/tomcat/tomcat-sam/logs/access.log*` + the snapshot copies | 6,722 requests from the XRAS caller |
| `/tomcat/tomcat-sam/logs/sam-xras-actions.log{,.2026-*}` | 211 processing lines, 2026-07-08 → 2026-08-07 |
| `/tomcat/tomcat-sam/webapps/ROOT/WEB-INF/` | `web.xml` + `javap` over `edu.ucar.cisl.sam.xras.rest.*Controller` |

---

## 1 · Production coverage is complete

Sole caller `18.223.62.77` (AWS us-east-2), UA `Ruby` — exactly as § 1.1 says.
Over 58 days, **four path shapes, nothing else**:

| Path — note the bare trailing `?` on every one | Hits | Status split | Ported |
|---|---:|---|:--:|
| `GET /api/xras/v1/people/{username}?` | 6,341 | 4,527 × 200 / 1,814 × 404 (**28.6%**) | ✅ |
| `POST /api/xras/v1/actions?` | 321 | 233 × 200 / 88 × 500 (**27.4%**) | ✅ |
| `GET /api/xras/v1/people?` | 59 | all 200 — one per night | ✅ |
| `GET /api/xras/v1/requests/request/{projcode}?` | 1 | 200, 2026-07-07 | ✅ |

Zero requests to anything else, and **no servlet-level 404/405 at all** — every
non-200 is an application 404 on `/people/{u}` or the known 500 on `/actions`.
So XRAS is not probing for endpoints we haven't mapped.

GETs are **95.2%** of the surface over the longer window, corroborating the 94%
in the description. The 28.6% `/people/{u}` 404 rate matches the runbook's
"~30% is the baseline, not an incident" exactly.

---

## 2 · The one gap: `/v1/roles/{requestNumber}/{role}/{username}` is unported

`javap` on `RoleServiceController`:

```
value=["/v1/roles/{requestNumber}/{role}/{username}"]     ← no method= → ALL verbs
```

This is a **write** — it reassigns the project lead. `XRAS_REIMPLEMENTATION.md`
§ 2.1 lists it as endpoint **#7** and describes it correctly. But it exists
nowhere else:

- no route in `src/webapp/api/xras/`
- absent from the blueprint docstring (`api/xras/__init__.py:11-18`, which lists
  six GETs and `POST /actions`)
- absent from this PR's description
- absent from `XRAS_CUTOVER_RUNBOOK.md` § *Known-open, and accepted*

Traffic in 58 days: **zero**. But cutover repoints one base URL, so after the
repoint this 404s where legacy 200s today, and nothing logs it. The asymmetry —
correctly documented in the design doc, invisible in every place an operator
would actually look — is the finding, more than the endpoint itself.

### The decompiled contract, for whoever ports it

`RoleServiceController` → `RoleServiceImpl` → `DefaultUpdateProjectLeadCommand`:

| Condition | Legacy exception | Status | Body |
|---|---|---:|---|
| `role` not `equalsIgnoreCase("pi")` | `NotFoundException` | **404** | `{"message":"role {role} does not exist","result":null}` |
| `@ProjcodeExists` fails | `NotFoundException` | **404** | `{"message":"non-existent project","result":null}` |
| `@ProjcodeActive` fails | `BadStateException` | **403** | `{"message":"inactive project","result":null}` |
| `@UsernameExists` fails | `BadStateException` | **403** | `{"message":"non-existent user","result":null}` |
| `@UsernameActive` fails | `BadStateException` | **403** | `{"message":"inactive user","result":null}` |
| any other validation message | `BadRequestException` | **400** | the raw message |
| success | `createOkResponse()` | **200** | ⚠️ **truly empty body**, no envelope |

⚠️ **403, not 400,** for three of the four failure modes —
`XrasController.handleBadStateException` maps to `HttpStatus.FORBIDDEN`. Easy to
get wrong from the ACCESS spec, which reads like these should be 404/400.

Four things the bytecode settles that reading the spec would not:

1. **The role check runs first**, before the project or the user is looked at —
   so a bad role against a nonexistent project is 404-*role*, not 404-*project*.
2. **`requestNumber` is used directly as the projcode.** `setLeadUserRole` passes
   it straight into `UpdateProjectLeadCommand.projcode(...)`. Consistent with the
   action log, where request number == projcode on every one of the 130 success
   lines.
3. **`transact()` does exactly two things**: `project.setProjectLead(user)` and
   `setModifiedTime(now)`. No roster insert, no allocation touch. That maps
   cleanly onto the existing `Project.update(project_lead_user_id=…)`
   (`src/sam/projects/projects.py:306`) — no new write primitive needed.
4. **The four messages are JSR-303 constraint templates**
   (`MessageResources_en_US.properties:79,80,86,87` — `@ProjcodeExists`,
   `@ProjcodeActive`, `@UsernameExists`, `@UsernameActive`) which `RoleServiceImpl`
   then *re-matches with a regex*. `String.matches` is a full-string match, so
   when **two** constraints fail at once the concatenated message matches none of
   the four patterns and legacy falls through to **400**. Reproducing that is
   optional; knowing it explains any stray 400 during triage week.

Note also that legacy maps **all verbs** (no `method=` in the annotation). The
ACCESS spec says POST. Mapping both POST and PUT is the same defensive move this
PR already makes for the two `/actions` URL forms.

### Related: nothing surfaces an unmapped path

An unmapped `/api/xras/*` path currently returns the blueprint's envelope 404 and
leaves no trace — which is precisely how #7 stayed invisible. A WARN log on
"no route matched under `/api/xras/`" turns "XRAS called something new" from a
silence into a grep, without moving any response bytes.

---

## 3 · `test_auth` is **not** a gap

`GET /api/xras/v1/test_auth` shows up 3× in the logs, but all three are from
`128.117.65.3` (UCAR VPN, `python-requests`/`curl` — our own probing), and legacy
answered **404**. § 2.1 already lists it as unimplemented-in-legacy. Correct as
written; no action.

---

## 4 · Trap worth adding next to the `grep -i xras` one

`54.85.201.121` (AWS **us-east-1**) also sends User-Agent `Ruby` — 3,248 requests
over the same window — but exclusively to `/api/protected/heuv/v1/*`. It is a
different system.

So filtering XRAS traffic by user-agent silently folds in a second integration;
filtering by **source IP**, as § 1.1 does, is right. Same class of mistake as the
AMIE `XRAS-*` task-name trap already documented, and worth sitting beside it.

---

## 5 · The error vocabulary is fully covered — verified against production

All **11** distinct strings legacy emitted over the last month map to a builder in
`src/sam/xras/errors.py`, byte for byte:

| Count | String | Builder |
|---:|---|---|
| 27 | `Could not determine Mnemonic code for internal PI via organization` | `mnemonic_internal_failed` |
| 18 | `Username {u} is missing` | `username_missing` |
| 18 | `PI {u} is not in database` | `pi_not_in_database` |
| 18 | `Could not produce affiliation data for PI {u}` | `no_affiliation_for_pi` |
| 18 | `Allocation Manager {u} is not in database: ` | `manager_not_in_database` |
| 6 | `Username {u} is inactive` | `username_inactive` |
| 6 | `PI {u} is not an active user: ` | `pi_not_active` |
| 6 | `Allocation Manager {u} is not active ` | `manager_not_active` |
| 6 | `Cannot find contract for grant number "{n}" ("{n}")` | `cannot_find_contract` |
| 5 | `Action end date is before existing allocation end date ({date})` | `extension_end_date_before_existing` |
| 1 | `Action end date before existing allocation end date for {resource}` | `update_end_date_before_existing` |

Including the trailing-colon and trailing-space typos, and including both end-date
phrasings — the two the module explicitly refuses to unify because the difference
tells an operator which path rejected the action. Production emits both, five days
apart. Good call.

---

## 6 · The silent-200 rate is measurable, and it argues for this PR harder than the description does

Over the window where both logs exist (2026-07-08 → 2026-08-07):

- access log: **241** posts to `/actions`
- `sam-xras-actions.log`: **211** lines

Per-day counts agree exactly on 20 of 22 days. The residual is **30 posts
(12.4%) that produced no record anywhere** — legacy returned 200 with a 41-byte
body and wrote nothing at all. On 2026-07-09 alone: 56 posts, 28 log lines.

That is exactly the hole `xras_action_log` + status `manual` + `outcome_reason`
closes, and it's a sharper number than "200, no trace" in the behaviour table.
The logs can't say how many of the 30 were non-serviceable vs. a duplicate
`actionId` — which is itself the argument for the `action_id` column.

---

## 7 · Two things in the description that are now stale

- **Test counts.** Body says "5,233 passed, 36 skipped, 1 xfailed";
  `docs/TESTING.md`, updated later on this same branch, says **5,309 collected /
  5,270 in the default run**. Updating the body.
- **"All 34 legacy error strings"** over-claims by three. `errors.py` has 34
  builders, but `ambiguous_role`, `adjustment_would_go_negative` and
  `ambiguous_contract` are documented *in the module* as deliberately **not**
  legacy strings. It's **31 legacy + 3 new** — and the body already explains why
  the 3 exist, so this is a number fix, not an argument.

Both corrected in the description, along with `/v1/roles` added to *What is left*.

---

### Summary

| | |
|---|---|
| Endpoints XRAS calls in production | 4 — **all ported** ✅ |
| Endpoints legacy maps | 8 — **7 ported**, `/v1/roles/…` outstanding ⚠️ |
| Production error strings | 11 — **all covered byte for byte** ✅ |
| Unmapped-path visibility | none today — recommend a WARN log ⚠️ |

PR #434, opened 2026-08-11
XRAS re-check: answer whether a failed action would succeed now

benkirk:
Replay could not answer the question it exists for. It re-ran `_parse_action` —
JSON well-formedness and the marshmallow schema — and stopped. Almost no real
failure is a schema failure: the ones operators chase come from the handler's
`assemble()` (organization and mnemonic resolution, contract lookup, allocation
type, roster, resource mapping). `New` succeeds ~30% of the time and the causes
are identity data, none of which the old check could see.

So a payload that would **still** be rejected came back `replayed`, indistinguishable
from one that would now succeed. An operator asking *"did my data fix take?"* got a
false yes on exactly the question they were asking.

## The loop this enables

**XRAS owns the retry**, so the only thing we control is whether their resend will
land. Now we can know before asking:

```
post fails 422 → fix the data here → re-check → green → ask XRAS to resend
                                             ↘ red  → the error list says what is still missing
```

## How

`dispatch_action(..., validate_only=True)` runs the handler's assemble-and-check half
and stops. That is a complete answer, not an approximation, because the template
already separates the write-free phase:

```python
self.assemble()              # contract: "Must not write."
self.errors.raise_if_any()   # → XrasActionRejected, before any transaction
with management_transaction(self.session):
    self.execute()
```

Nothing can be written on that path — a property of the template rather than of a
flag, so it still holds when `XRAS_ACTIONS_CAPTURE_ONLY` flips at cutover. Outcomes
reuse the ingest vocabulary, so a re-check row reads like the ingest row it would
have been: `rechecked` (would succeed), `failed` + the real error list (would still
fail), `manual` (nothing would run).

Two things the tests caught, both worth reading:

- **`run(validate_only=True)` must not call `result()`.** Subclasses override it to
  report execution state — `ExtensionHandler.result` dereferences `self.extended`,
  which only `execute` creates. The four common fields are restated instead, so a
  subclass adding a post-commit report cannot break re-checking.
- **`projcode_result` stays NULL.** A re-check produces no project, and on the New
  path the handler's `projcode` is the *request token* (`NCAR4253`) — writing it
  would put a non-projcode into the column that joins `project.projcode`.

## The rename, while the window is open

All three production tables are empty and XRAS has not repointed, so the naming can
still be fixed. That window closes permanently at cutover.

| | From | To |
|---|---|---|
| Column | `replay_of_id` | `source_action_id` |
| Status | `replayed` | `rechecked` |
| FK + index | `xras_action_log_replay_fk` | `xras_action_log_source_fk` |
| Relationship | `replay_of` / `replays` | `source_action` / `rechecks` |
| Module + function | `replay.py` / `replay_action` | `recheck.py` / `recheck_action` |
| Route | `xras_replay` | `xras_recheck` |
| CLI | `--replay` | `--recheck` |

"Replay" implied re-execution — exactly what this does not do, and the
mis-implication that made a safe tool read as dangerous. `source_action_id` is
neutral, so a future re-apply path can share the lineage column.

## ⚠️ Deploying this needs one manual step, in a specific order

**Production is deliberately NOT yet altered.** The snapshot was regenerated by
cloning production, altering the *local* copy, then dumping — `make bootstrap`
cannot do it unattended, since it re-clones and would discard the change.

At deploy time, **ALTER production immediately before dispatching the rollout**:

```sql
ALTER TABLE xras_action_log RENAME COLUMN replay_of_id TO source_action_id;
ALTER TABLE xras_action_log
  DROP FOREIGN KEY xras_action_log_replay_fk,
  ADD CONSTRAINT xras_action_log_source_fk
    FOREIGN KEY (source_action_id) REFERENCES xras_action_log (xras_action_log_id);
ALTER TABLE xras_action_log RENAME INDEX xras_action_log_replay_fk TO xras_action_log_source_fk;
```

Both orders leave a window — old and new pods coexist during a rolling update — but
they break different cohorts. ALTER-first breaks the pods being drained anyway and
self-heals when the last new pod is Ready; rollout-first breaks the pods that stay,
and leaves prod broken until someone remembers the SQL.

ALTER-first is safe specifically because the k8s probes use `/live` and `/ready`,
which exclude the schema-drift check by design — so drift cannot mark pods unready
or empty the Service. Only `/api/v1/health/` reports 503, which is monitoring-visible
and harmless to k8s. Affected surfaces are the XRAS tab, the action-details modal and
`POST /actions`; the rest of the app never touches this column. Rollback is the
reverse ALTER on an empty table.

## Also

- The control is offered only for `failed` / `manual` / `rechecked` rows. Re-checking
  a `processed` row asks about a world that no longer exists: it already changed the
  data it would run against, so a successful New routes to Update and the answer
  describes a different action.
- The route reports the **verdict** ("Would succeed now") rather than the mechanism
  ("Replayed action #N"), and no longer interpolates an exception into the page.
- Registry contract is now `(session, action, *, validate_only=False) -> DispatchResult`.

## Snapshot

Proved by dropping the test volume and restoring from the blob alone — it returns
with `source_action_id`, the renamed FK and all 7 utf8mb4 columns, so the rename
travels in the snapshot rather than in a local session. 3 tables, 0 rows, 28,372
obfuscated usernames, 2 real addresses (both `preserve_usernames` entries).

⚠️ Correcting the previous blob's write-up: it claimed every phone number is
`555-`-prefixed. Twelve are not — they take the international branch of
`_get_fake_phone`, which reuses one `hash_val` per group, giving shapes like
`860-860-2860`. Fully synthetic and retaining none of the original digits, but the
earlier check sorted by frequency and printed only the top entries, hiding
singletons. The check no longer sorts.

## Test Results

```
5,678 passed, 36 skipped, 1 xfailed   (default, ~95s)
   21 passed                          (stress)
   22 passed                          (perf)
```

Four new tests, including the one that would have caught the original defect (a
handler that still rejects must land `failed` with its reasons, not `rechecked`) and
one that boobytraps `execute` to prove the write path is unreachable under
`CAPTURE_ONLY=False`. Route-map snapshot regenerated: one endpoint renamed, two lines.


benkirk:
Superseded by a branch based on `staging` rather than `integration`.

#427 was squash-merged, so `staging` holds integration's *content* under a single new commit while `integration` kept its own 11-commit history. A branch off `integration` therefore replayed all 11 already-merged commits at `staging` — hence the conflicts, and hence no CI: GitHub cannot build the `refs/pull/N/merge` ref for a conflicting PR, so `pull_request` workflows never fire.

Same two commits, cherry-picked onto `staging`. Tree is byte-identical; the diff drops from 227 files to 34.

PR #453, opened 2026-08-17
Cache /static for a year, with content-hash cache busting

benkirk:
## The problem, measured

Flask leaves `SEND_FILE_MAX_AGE_DEFAULT` at None, which sends `Cache-Control: no-cache` on every static file. Browsers cache correctly under that — they just have to **revalidate on every page load**, so each asset costs a conditional round trip for ever.

Measured on nwc1 2026-08-17, one webapp pod, a 4h37m window (07:28–12:05 MDT):

| | |
|---|---|
| `/static` share of logged requests | **4,336 of 4,953 — 87.5%** |
| of those, answering `304 Not Modified` | **4,152 — 96%** |
| vendored `bootstrap-5.3.3` revalidations | ~145 in 4½ hours |

A 304 carries no body, but it's still a full request through nginx, gunicorn and Flask. Those 304s were also the dominant consumer of `max_requests` in `gunicorn_config.py` — whose entire purpose is to bound *application* heap growth — so the worker-recycle clock was being driven almost entirely by requests that touch nothing on the Python heap.

**Provenance note:** this surfaced while chasing the 2026-08-17 cron connect timeouts (fixed client-side in `NCAR/hpc-scheduling-tools#12`). Worker churn was the first suspect and turned out to be entirely planned recycling — 16 `Autorestarting` = 16 `Booting worker`, **zero** `WORKER TIMEOUT`, **zero** `SIGKILL` across a 20k-line window. `max_requests = 1000` is a placeholder from the file's initial commit that has never been revisited, and this PR **deliberately leaves it alone**: its measured cost is one graceful fork per ~31 min with memory at a fifth of the limit, and raising it would unmask a leak nobody has characterized, for no benefit. The traffic *mix* was the real finding.

## The fix

Two halves, only correct together:

- `url_for('static', filename=...)` grows `?v=<content hash>` — a changed file gets a new URL
- a response carrying that parameter is declared immutable for a year

**ONE rule sets the header** — is `?v=` present:

| request | `Cache-Control` |
|---|---|
| carries `?v=` | `public, max-age=31536000, immutable` |
| no `?v=` | `public, max-age=3600` |

Before choosing that shape I verified the preconditions: **38** `url_for('static')` calls in templates, **zero** hardcoded `/static/` paths in templates, and **zero** `/static/` references inside the CSS/JS. So the immutable branch covers 100% of the measured traffic.

The short branch is load-bearing, not a fallback — it covers the assets no `url_for` can reach: relative `url()` targets inside a stylesheet (FontAwesome's `../webfonts/*`, the one `../img/` in the app's own CSS). Those get a bounded TTL instead of an unbounded one, so replacing such a file *in place* can't pin a stale copy in a browser for a year.

## Review notes

- **No vendor path special-case, on purpose.** `vendor/bootstrap-5.3.3/` and `vendor/fontawesome-6.5.2/` carry their version in the path and would be safe to hardcode as immutable — but a path table is a second thing to keep in sync with the tree layout, and it buys nothing the `?v=` rule doesn't already give. The rule is shaped so the **conservative branch is the default**: getting it wrong costs efficiency, never correctness.
- **Content hash, not mtime.** A container rebuild restamps every mtime, so an mtime tag would invalidate every asset on every deploy — safe, but it would discard exactly the cache this creates. Asserted from both directions: changing content changes the tag, `touch` does not.
- **`asset_version` returns None rather than raising** on an unreadable file. A template naming a missing asset is already a 404 the browser reports; a 500 on every page that mentions it would be strictly worse. Reads through werkzeug's `safe_join` — the one place a stray `..` in a `url_for` kwarg would escape the static root.
- **Header is assigned, not `setdefault`'d** (unlike the security headers): the value being replaced is Flask's own `no-cache`, and only the static endpoint reaches this hook. `Expires` is dropped — `send_file` sets it alongside its own `Cache-Control`, and a leftover value would be obeyed in preference to ours by an HTTP/1.0 cache.
- **18 new tests**, including the end-to-end one that subsumes the rest: *a real rendered page must emit only versioned static URLs*. That's the assertion that would have caught the original problem, rather than the mechanism that fixes it. Plus two lint gates — no template may hardcode `/static/` (efficiency), and none may append `?`/`#` after `url_for('static')` (correctness: the URL now already carries a query string, so `{{ ... }}?x` yields `?v=abc?x`).

## Interaction worth knowing

Rendered HTML is itself cached (`user_aware_cache_key`), so after a deploy that changes an asset, cached fragments keep emitting the **old** `?v=` until they expire, and a browser holding that URL keeps using its cached copy. Bounded by the HTML cache TTL and already covered by the documented post-deploy step, `sam-admin cache --refresh`. The asset is never unreachable — `v` is a cache key, not a lookup key, so any value serves the current file.

**Not attempted here:** moving `/static` to the ingress. That would cut these requests to zero rather than ~⅛, but it needs the assets alongside the controller and is a chart change. The header fix gets most of the win with no infrastructure move.

## Test Results

Full default suite green: **7,051 passed, 42 skipped, 1 xfailed** (89s).

`test_vendor_assets.py` asserts `f"/static/{path}" in html` — a substring check, so it survives the added query string unchanged. CSP tests are unaffected because CSP URL matching ignores query strings.

Verified end-to-end on a rendered page: every static reference on `/auth/login` carries `?v=`, and fetching one returns `public, max-age=31536000, immutable` with no `Expires`.

## After deploy

Watch `/static`'s share of requests drop from ~87% toward ~10%, and the pod-wide worker-recycle interval stretch well past its measured ~31 min — same `max_requests`, far fewer requests reaching it. Run `sam-admin cache --refresh` as usual so cached HTML picks up the new tags immediately.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -409,6 +409,48 @@ Template side: emit `value="1"` and wire both directions via the
 card switch, use `hx-trigger="change"` + `hx-include="this"` (see the
 Resources/Organizations/Facilities cards).
 
+### 11. Static Assets & Cache Headers
+
+**Always reference an asset with `url_for('static', filename=...)`** — never a
+literal `/static/...` path. `webapp/utils/static_cache.py` appends
+`?v=<content hash>` to every `url_for` static URL, and that parameter is what
+buys the asset a one-year cache.
+
+Flask's default is `Cache-Control: no-cache`, which makes a browser revalidate
+every asset on every page load. Measured on nwc1 2026-08-17 (one pod, 4h37m):
+**87.5% of requests were `/static`, and 96% of those answered 304** — body-less
+round trips whose only purpose was to say "unchanged". Vendored Bootstrap was
+revalidated ~145 times in 4½ hours. Those 304s were also the dominant consumer
+of `max_requests` in `gunicorn_config.py`, whose point is to bound *application*
+heap growth.
+
+**ONE rule sets the header** — is `?v=` present:
+
+| request | `Cache-Control` |
+|---|---|
+| carries `?v=` | `public, max-age=31536000, immutable` |
+| no `?v=` | `public, max-age=3600` |
+
+No path table, no vendor special case. The conservative branch is deliberately
+the **default**, so an asset referenced some novel way degrades to a short TTL
+rather than pinning a stale copy for a year — getting the rule wrong costs
+efficiency, never correctness. The short branch is load-bearing for the assets
+no `url_for` can reach: relative `url()` targets inside a stylesheet
+(FontAwesome's `../webfonts/*`, the one `../img/` in the app's own CSS).
+
+- **Content hash, not mtime.** A container rebuild restamps every mtime, so an
+  mtime tag would invalidate every asset on every deploy — safe, but it would
+  discard exactly the cache this exists to create.
+- ⚠️ **Rendered HTML is itself cached** (`user_aware_cache_key`), so after a
+  deploy that changes an asset, cached fragments keep emitting the old `?v=`
+  until they expire. Covered by the existing post-deploy step,
+  `sam-admin cache --refresh`. The asset is never unreachable — `v` is a cache
+  key, not a lookup key.
+- Gates in `tests/unit/test_static_cache.py`: a real rendered page must emit
+  **only** versioned static URLs, no template may hardcode `/static/`, and no
+  template may append `?`/`#` after `url_for('static')` (that one is a
+  correctness bug — the URL already carries a query string).
+
 ---
 
 ## Flask-Admin (`src/webapp/admin/`)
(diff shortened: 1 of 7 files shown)


benkirk:
Second commit added after a review question: *does this sit beside the existing cache framework intentionally, or is that context blindness?*

Honest answer was **half and half**, so both halves are now addressed.

**The name was doing the confusing.** `webapp/utils/static_cache.py` sat one directory from `webapp/caching/`, whose own docstring claims to be *"the single entry point for every cache layer in the webapp"*. Renamed to `static_assets.py` / `init_static_assets()` — names the subject rather than the mechanism. No behavior change; same header rule, same `?v=`, same 18 tests, suite still 7,051 green.

**The exclusion is now argued in the code, not assumed**, because it isn't one-sided:

- The **response header** genuinely isn't a cache this app owns — nothing stored, no entry to count, and `caching.clear()` cannot reach a browser cache. The dependency runs the other way and is real: the remedy for cached HTML holding a stale `?v=` is `caching.clear('flask')`. It's an `after_request` header shaper, which is why it belongs beside `security_headers.py`.
- The **digest memo** inside `init_static_assets` *is* a server-side store, and it's the only bespoke memo in the webapp outside the facade. It stays out deliberately: `BucketedTTLCache` is shaped for derived data whose inputs change — TTL, category, admin card row, working `clear()` — and this memo wants the inverse of all four. File bytes cannot change inside a running container image, so the correct TTL is infinite, a `clear()` could only force pointless re-hashing, and size is bounded by file count rather than traffic.

The docstring labels that second bullet a narrow exception **with stated reasons and an explicit disclaimer of precedent**, so the next bespoke memo can't cite this one on its way out of the facade. A comment at the memo repeats it for anyone reading only the code.

Both commits are recorded as renames, so the diff stays readable.

Issue #433, opened 2026-08-11
XRAS → SAM reintegration: endpoint cutover plan


benkirk:
NSF NCAR is moving SAM — our allocation and accounting system — from Java to
Python/Flask. The endpoints XRAS calls have been reimplemented and are running in
production alongside the originals.

**For XRAS this is a one-line change: repoint the base URL.**

| | |
|---|---|
| Today | `https://sam.ucar.edu` |
| After | `https://sam.hpc.ucar.edu` |

Everything after the hostname is unchanged — same paths, same request formats, same
JSON responses, and the **same Basic-auth credential**.

### How we verified it

The new stack has been compared against the live legacy stack **byte for byte** on the `GET` endpoints, using
the same credential against both hosts.  We've tested a number of old `JSON` payloads for the `PUT`s but this is
probably our biggest risk, missed edge cases, and we'll be prepared to triage incoming at deployment.

---

## Endpoints

Paths are unchanged; only the hostname moves. "Notes" records inherited quirks we
reproduced deliberately — see *Candidates for a v2* below.

| Endpoint | Change | Notes |
|---|---|---|
| `GET /api/xras/v1/people` | none | **No envelope.** Returns all 28,259 users including inactive; ~3.84 MB, ~1.1 s per call; no filter or paging |
| `GET /api/xras/v1/people/{username}` | none | **No envelope.** ~30% of lookups 404 by design — historical usernames, not an error rate |
| `GET /api/xras/v1/requests/request/{requestNumber}` | none | Enveloped |
| `GET /api/xras/v1/requests/user/{username}` | none | Enveloped |
| `GET /api/xras/v1/requests/role/{role}/{username}` | bogus `{role}` → **`400`** with a message, was `500` | Enveloped. Valid roles: `pi`, `co_pi`, `allocation_manager` |
| `GET /api/xras/v1/dates/requests/{requestNumbers}` | none | Enveloped. Takes a comma-separated list inside a **path segment** |
| `POST /api/xras/v1/actions` | **`400` / `422` / `200` now distinguish outcomes** — see below | `Transfer` actions park for manual handling; zero traffic to date |
| `POST /api/xras/v1/roles/{requestNumber}/{role}/{username}` | **`404`** (not found) and **`409`** (inactive) replace a blanket `400` | Only `role=pi` is accepted; any other role returns `404` |

An unrecognized path under `/api/xras/` now returns a JSON 404 envelope rather than a
servlet-container HTML error page.

[...]

PR #457, opened 2026-08-19
XRAS cutover: release the capture-only interlock, and a triage playbook

benkirk:
## Summary

Closes out [#433](https://github.com/benkirk/sam-queries/issues/433). Four commits: one behaviour change, one config flip, and the two documents triage week will actually be run from.

Verified against the live systems on 2026-08-19, before any of this:

| | State |
|---|---|
| `sam.hpc.ucar.edu` | Real InCommon cert, SANs for both hosts, valid to 2027-02-13. Serves `/api/xras/v1/*`, answers the 41-byte 401 unauthenticated |
| Production | `sha-6e05e29`, `XRAS_ACTIONS_CAPTURE_ONLY=1`, `NOTIFY_ENABLED=1`, `NOTIFY_TRANSPORT=smtp` |
| `xras_action_log` on prod | **0 rows, every status** — so XRAS has not repointed, nothing is stranded, and nothing can have been double-applied |
| Runbook gates 1, 2, 4 | Closed |

## 1 · Tell XRAS when an action was parked rather than applied

`processed` and `manual` both answered `{"message":"OK","result":null}`, byte for byte. So an admin who pushes the button in `xras_admin` for an action SAM does not service is told it worked.

Legacy does the same, which is why this shipped as parity. It stopped being a corner case when `Date Adjustment` was discovered on 2026-08-11: it parks, and it is 4 of the 41 corpus payloads. `Transfer` parks by design too. The silent-success arm is the **common** one.

Steven Peckins asked for exactly this on the thread that settled the 400/422 contract — *"the response body is saved and made available in xras_admin for the admin to see, so it's nice to include something informative"*. Both halves of his reply are load-bearing:

- **The status stays 200.** *"Any status other than 200/OK is considered an error."* A parked action is not one; 202 would have been the tidy REST answer and would have read as a failure.
- **Only the `manual` body moves.** `processed` still answers exactly `OK`, and so does the capture-only arm.

The message is a module constant, deliberately *not* `DispatchResult.reason` — three of the four parking causes name internal machinery (`XRAS_ACTIONS_ENABLED`, an unregistered handler) an ACCESS admin can neither act on nor should see. That detail keeps going to `xras_action_log.outcome_reason`, which is what our own surfaces read.

This is gate 4's open decision, now closed.

## 2 · Release the capture-only interlock

`XRAS_ACTIONS_CAPTURE_ONLY: "1"` → `"0"`.

**Deliberately ahead of the repoint, not in step with it.** The runbook read as if the two steps were simultaneous. They are not symmetric, and the asymmetry runs opposite to the intuition that a safety interlock comes off last:

- **Early costs nothing.** XRAS still posts to `sam.ucar.edu`, so the dispatching arm sees no traffic until the repoint.
- **Late costs a round-trip with ACCESS, per action.** Posts in the gap are stranded as `status='received'`, and `--recheck` **cannot apply one** — `dispatch_action(..., validate_only=True)` returns before `management_transaction` opens, structurally. Recovery is asking the XRAS admin to push the button again for each.

The thing that must not happen is a flip on a stack XRAS is *already* posting to, which double-applies against live allocations with no undo. So it is checked rather than assumed: the summary above is that check, and it is clean.

`XRAS_ACTIONS_ENABLED` stays `"all"`. `xras_notices` stays in `SAM_TASKS_DISABLED` for triage week, so handoff notices go out from the Notify button with a human looking at each one; clearing it is a follow-up after the week.

## 3 · The triage playbook

`docs/xras/incoming/XRAS_TRIAGE_PLAYBOOK.md`. The runbook ends at the repoint and gives triage four bullets; from Monday the question is *"this action did not land, now what"*, and that answer was spread across `errors.py` docstrings and three sprint records.

- **The daily loop**, and what normal looks like — the ~30% `/people` 404 baseline, the 3.84 MB roster, `New`'s historical ~30% success. The numbers that otherwise get chased on day one.
- **Classify before you fix**: `status` × `http_status` × `service` × `outcome_reason` as one table.
- **The 422 catalog**, ranked, each with its data fix. The top three are all data, not code: unmapped resource keys (13 mapping rows exist, 11 active resources have none), the organization mnemonic gap (153 of 171 orgs — 24% of legacy's failures), and unreconciled ARC identities (55% of `New` failures).
- **Levers**, and what each costs.

Two things it insists on, because both are easy to get wrong under pressure:

- **`--recheck` validates, it cannot apply.** Green means "this would work if XRAS sent it again". Every fix ends by asking ACCESS to re-post, and is not done until the action comes back and lands.
- **Adding a resource mapping moves GET response bytes**, because `resourceRepositoryKey` is omitted when a resource has no row. The most obvious triage-week fix silently invalidates the gate-3 parity run, so the re-run command sits next to the fix.

**§ 6 sketches five remediation tools and builds none of them**, naming the *trigger* for each instead — including the re-apply path, which needs an idempotency key on `action_id` first because 4 of the 6 handlers double-apply. Writing remediation code for failures that have not happened is how you end up maintaining the wrong tool.

## 4 · Runbook housekeeping

- **Stale helm line refs.** The runbook cited `values.yaml:291`/`:295`; the settings are at **316**/**320**, and 291/295 now hold fs-scans comments. Cutover day is exactly when someone follows a line number without reading around it.
- Gate 4 recorded as decided, gate 5 recorded with the asymmetry above, playbook linked.
- A `values.yaml` comment still listed `deactivate_expired_projects` as disabled after #455 turned it on.

## Test Results

```
pytest -q                      7054 passed, 42 skipped, 1 xfailed   (83s)
pytest -m stress -n 0            22 passed
tests/api/test_xras_access.py   155 passed
helm lint                         0 failed
```

Three new tests in `TestDispatchArms` pin what moved and what did not: parked is not byte-identical to a success, the parked body does not leak `outcome_reason`, and a processed action still answers exactly `OK`.

---

# Rebased onto staging, 2026-08-20 — three more commits

#458 and #459 both merged after this branch opened. Rebased so triage week rides on the combined capability, and because a `workflow_dispatch` from a stale branch would revert their chart changes (below).

## ⚠️ The interlock flip is currently **reverted in production**

`origin/cirrus` is `68cf9b9`, dispatched from `xras_opportunityId`, and its `helm/values.yaml` is staging's — so `XRAS_ACTIONS_CAPTURE_ONLY` is back to `"1"`. That deploy force-pushed its own branch's chart over this branch's flip. It is the trap the note below already warned about, and it has now fired: the `update-helm` job checks out `ref: ${{ github.ref }}` and pushes *that ref's* chart, so whichever branch dispatched last wins.

**Harmless as of now** — `xras_action_log` is still 0 rows, ACCESS has not repointed, so nothing was stranded and nothing double-applied. But the "confirm `CAPTURE_ONLY=0` in the pod env" box below described a state that no longer holds, and a redeploy from this branch is what restores it.

## 5 · Chart: the task roster after #455, #458 and #459

Three comment corrections plus one missing knob.

The Live/Disabled roster still said *"exactly one task is live"* and still listed `deactivate_expired_projects` as disabled after #455 enabled it — and never gained `xras_sweep` after #458 shipped it on. Three tasks are live. That comment is the only human-readable index of what runs in production, because `SAM_TASKS_DISABLED` is fail-OPEN and names only the exceptions.

⚠️ **`xras_sweep` is no longer read-only and its own comment said it was.** #458 wrote *"writes nothing but a cache entry the dashboard reads"*; #459 made it insert `xras_opportunity_allocation_type` rows — what `xras_sweep.py` itself calls "the only write this task performs".

`SAM_TASKS_XRAS_MAP_MAX` was the one sweep knob the chart never named, so the cluster silently ran the code default while `MAX_PAGES` / `MAX_PEOPLE` / `WINDOW_DAYS` / `STATUS` were all pinned beside it. Pinned at that default — behaviourally a no-op — and confirmed rendering into the CronJob.

## 6 · `sam-admin xras --validate-opportunities`

The last deferred piece of #459, and the exception to the playbook's "sketch it, don't build it" rule. The reason is the failure *mode*, not the likelihood.

Every other XRAS mapping gap shouts: an unmapped `resourceRepositoryKey` 422s the action. An unmapped `opportunityId` does not — it falls through to the free-text ladder, whose twelve `SelectionParms` pairs never name `UW`, `WRAP` or `LCAP`, so a Wyoming `Small` request resolves to panel `UNIV USS`, the join **succeeds**, and the project gets a UNIV-series projcode. Nothing fails. **So the trigger cannot arrive as a symptom** — there is no row to classify, and the check has to run against the map ahead of the action.

CLI wiring over machinery #459 already shipped (`audit_opportunity_mapping`, `propose_opportunity_mapping`, `get_open_opportunities`). Three decisions:

- **Exit non-zero on `dangling_ids` only.** Not `unmapped_ids` — with an empty table every opportunity is unmapped and ingestion is perfectly healthy. Not `review` — two pairs sit there permanently by design, and a bucket that is never empty is a bucket an operator stops reading. Same trap `--validate-mapping` fell into once and corrected.
- **The proposal covers the unmapped subset only**, mirroring `_map_new_opportunities`, or the four `source='manual'` rows reappear in `review` every run.
- ⚠️ **It reads the OPEN list, not `/v1/opportunities/list/:ids`** — the opposite of what the design doc specified. That instruction is right for backfill and wrong for a check: the historical tail is `xras_sweep`'s job and those ids are closed, while the open list is the only place a brand-new opportunity appears. Recorded in § 8.6 of the design doc rather than left as a silent divergence.

Both guards negative-tested: widening the proposal fails exactly one test, exiting non-zero on `unmapped_ids` fails five.

## 7 · Reconciling the playbook with what now exists

The playbook was written the day before #458/#459 landed.

| | |
|---|---|
| **One sentence was flatly wrong** | § 3.1 said dangling keys were "the only one that makes the command exit non-zero". `--validate-mapping` is two-sided since #458 and also exits non-zero on `xras_only_keys` |
| **The #1 ranked 422 is demoted** | "13 mapping rows, 11 active SAM resources have none" is still true, but those are opposite directions and only one can cause the 422. #458 measured the one that matters: **13/13 of the keys XRAS actually offers resolve.** It now needs XRAS to offer a resource we have never seen |
| **The largest cause got a queue** | 55% of `New` failures had the fix "identity reconciliation" — a category. It is now the Accounts Needed tab and `--accounts`, documented with the three traps: `placeholder` is a flag not a classification, `isReconciled` says nothing about SAM, `isAccountToBeCreated` is never the predicate |
| **§ 3.9 is new** | The two failures that appear in *none* of `status`/`http_status`/`service`/`outcome_reason`, because both present as success: a wrong opportunity map row (silent at ingest — the symptom is a `processed` action with the wrong projcode series), and `published: true` with `publish_backend != redis` |
| **The pre-repoint empty state** | A populated Pending Requests tab beside an empty Accounts Needed tab is **correct**. Without that line, day one goes on debugging a designed empty state |

✅ **Gate 3 does not need re-running** — checked, not assumed, because for two large XRAS PRs the assumption runs the other way. `webapp/api/xras/` is untouched by both; `roster.py`'s nine new lines are a comment block; neither adds a resource-mapping row; the opportunity map is read only by the inbound POST path. Recorded in the runbook so nobody burns the job on a hunch, and so the real trigger stays sharp.

⚠️ **The `values.yaml` line refs were wrong again.** `8dd454d` repaired `:291` → `:316` — against the *base*, while the same branch's comment rewrite pushed the settings down sixteen lines. Now cited by key name, with a note saying why, so they do not come back a third time.

## Test Results — rebased

```
pytest -q                          7338 passed, 42 skipped, 1 xfailed  (94s)
pytest -m stress -n 0                22 passed
tests/unit/test_admin_xras_cli.py    52 passed  (9 new)
helm lint + test-cronjob-render.sh    0 failed
```

Exercised by hand against the live API from the test database, all three arms: lever off → local half only, exit 0; unreachable host → yellow warning, local half, exit 0; configured → 6 open opportunities, 2 unmapped, both agreeing, exit 0. That last one is real content — `Large Allocation (University) - Fall 2026` is an opportunity the map does not have yet and the sweep will write.

---

## Before merging / deploying

- [x] **Gate 3 parity** — passed 2026-08-19, 13/13 byte-identical, and **not invalidated** by #458/#459 (§ 7).
- [x] Re-run the empty-log check immediately before the deploy: `sam-admin xras --summary` → **0 rows, every status**, so the early flip was safe.
- [x] **Redeploy CIRRUS from this branch** → `sha-82d8c00`, 2026-08-20 19:04 MDT.
- [x] After deploy: `XRAS_ACTIONS_CAPTURE_ONLY=0` in both pods **and parsed `False` by the app**; mailer verified live via `NotifyConfig.from_environment()` (not `app.config`, which misreads — see the validation comment below).

⚠️ A `workflow_dispatch` from this branch force-pushes **this branch's** `helm/values.yaml` onto `cirrus`. Keep it rebased on `staging` for the duration, or a dispatch reverts anything merged to staging since — the trap #367 hit, and the trap that just cost this branch its flip.

Issue #433 is also stale and worth updating before the call: it still presents the broker-retry question as *"the only thing blocking us"*, but Steve answered it on 2026-08-11.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,15 @@ megalinter-reports/
 .coverage.*
 Vermilion_NSFNCAR-UCAR-UCP_BrandGuide.pdf
 .playwright-mcp/
+# ⚠️ Playwright/MCP writes snapshots and screenshots to the CWD when a relative
+# filename is passed, not into .playwright-mcp/ — so they land in the repo root.
+# Smoking the XRAS cards means rendering UNSCRUBBED production PII, and two such
+# files (a page snapshot and a full-page PNG) were caught one `git add` away
+# from a public repo. Ignore them by shape rather than relying on the reviewer.
+page-????-??-??T*.yml
+*-pane.md
+*-before.png
+*-after.png
 legacy_sam
 data/
 
--- a/compose.yaml
+++ b/compose.yaml
@@ -49,6 +49,12 @@ services:
       #    fall back to compose service-name defaults. If the user
       #    uncomments PROD in .env, that flows through to in-stack
       #    containers (intentional pass-through).
+      # ⚠️ Match the chart (helm/values.yaml sets TZ: America/Denver). SAM's
+      # MySQL columns are naive-MOUNTAIN, stamped from the app clock — so a
+      # container left on UTC writes `received_time` six hours in the future
+      # relative to the data around it, and any column that renders an age
+      # comes out negative. Prod has always set this; local never did.
+      - "TZ=${TZ:-America/Denver}"
       - "SAM_DB_SERVER=${SAM_DB_SERVER:-mysql}"
       - "SAM_DB_USERNAME=${SAM_DB_USERNAME:-root}"
       - "SAM_DB_PASSWORD=${SAM_DB_PASSWORD:-root}"
@@ -134,6 +140,12 @@ services:
     # when adding new files/functions. Remove for production deployment.
     environment:
       # ── DB routing (see `webapp` for the pass-through reasoning)
+      # ⚠️ Match the chart (helm/values.yaml sets TZ: America/Denver). SAM's
+      # MySQL columns are naive-MOUNTAIN, stamped from the app clock — so a
+      # container left on UTC writes `received_time` six hours in the future
+      # relative to the data around it, and any column that renders an age
+      # comes out negative. Prod has always set this; local never did.
+      - "TZ=${TZ:-America/Denver}"
       - "SAM_DB_SERVER=${SAM_DB_SERVER:-mysql}"
       - "SAM_DB_USERNAME=${SAM_DB_USERNAME:-root}"
       - "SAM_DB_PASSWORD=${SAM_DB_PASSWORD:-root}"
--- a/docs/plans/XRAS_ACCOUNT_QUEUE.md
+++ b/docs/plans/XRAS_ACCOUNT_QUEUE.md
@@ -1,0 +1,118 @@
+# The XRAS account queue — deferred work
+
+**Status: written down, deliberately unbuilt.** Everything here was scoped during the
+2026-08-20 pass that made the worklist legible, and cut from it on purpose so triage week
+ships without new tables or new mail. Each item names what it is, why it was deferred, and
+what would justify picking it up.
+
+Context first: [`../xras/PROJECT_AND_ACCOUNT_LIFECYCLE.md`](../xras/PROJECT_AND_ACCOUNT_LIFECYCLE.md)
+— in particular **SAM never creates users**, which is what makes this a queue handed to
+another team rather than a control surface.
+
+---
+
+## 1 · `xras_account_event` — making it a *worked* queue
+
+**The blocking item.** The card is read-only. Two people working it cannot see each
+other, "asked upstream on the 14th" cannot be recorded, and a row that should not be
+worked at all cannot be set aside.
+
+Already designed at `../xras/outgoing/XRAS_OUTGOING_QUERIES.md` § 7.6. The shape:
+username-keyed, append-only, `event_type` from a small tuple, `comment`, `created_by`,
+`xras_action_log_id` as provenance only.
+
+⚠️ **`XrasActivationEvent` cannot carry it.** Its `project_id` is NOT NULL and
+project-scoped; this worklist is username-keyed and a New request *has no project yet*.
+That is the whole reason for a second table.
+
+⚠️ **Derive state with the action-keyed supersedes idiom in
+`sam/queries/xras_activation.py::_activation_state`** — not the older project-keyed rule
+quoted in the `XrasActivationEvent` docstring.
+
+**The property that keeps it small: only dismissal needs storage.** "Done" is already
+derived — classification is a check against the current `users` table on every render, so
+a row vanishes when the mirror lands and nobody has to close anything. This is a
+note-and-hide table, not a ticket system. Suggested vocabulary: `requested` (we asked
+upstream, with the date), `comment`, `dismissed`, `restored`. Deliberately **no**
+`created` — SAM cannot observe that except by the row disappearing.
+
+DDL applied by hand and recorded in `../xras/incoming/XRAS_CUTOVER_RUNBOOK.md` § 2,
+matching the precedent set by the other XRAS tables; ORM follows the database. Write path
+copies the Pending Activations routes (`webapp/dashboards/allocations/blueprint.py`,
+the `xras_activate` / `xras_dismiss` / `xras_restore` / `xras_comment` family), which
+already carry CSRF, `management_transaction` and the `HX-Trigger` refresh.
+
+**Trigger:** two people working the queue at once, or the first time somebody asks "did
+we already ask about this person?"
+
+---
+
+## 2 · A digest, so nobody has to poll a tab
+
+The audience already fields mail — `webapp/utils/rbac.py:258` records that the XRAS
+failure mail goes to `hdt@ucar.edu` today. A digest replaces mail with mail, which is the
+lowest-friction change for the people receiving it.
+
+~80% of this is copy-paste from `src/scheduling/tasks/expiration_notices.py`: the guards
+(`NotificationsDisabled`, `EmailCapExceeded` from `scheduling/tasks/mail_guards.py`), a
+per-run `SAM_TASKS_*_MAX` cap, the `_drop_already_notified` pre-filter, the own-session
+ledger, `requested_by='task:<name>'`, and `detail` always carrying
+selected/suppressed/audience.
+
+⚠️ **The one genuinely open design question is the dedup key**, because a *snapshot*
+digest matches neither existing shape. `expiration_notices` keys on something that
+changes when the fact changes (the expiration date); `xras_notices` keys on the action id.
+A worklist row persists week over week:
+
+- key on the **week** → an unchanged list re-sends every week (probably right for a digest)
+- key on the **username** → a person is reported exactly once, ever (almost certainly wrong)
+
+**Recommendation:** key on the **occurrence**, as `task_summary` does, and let the body
+carry the delta — *"3 new since last week, 9 still waiting, oldest 37 days"*.
+`sam/notify/templates/task_summary.txt` is already the closest body shape.
+
+Also needed: a new `NotificationKind` (`Notifier` hard-raises on an unregistered kind,
+`sam/notify/kinds.py:150`) plus `{base}.txt` / `{base}.html`, `facility_aware=False`.
+
+⚠️ **Ship it named in `SAM_TASKS_DISABLED` in the same change** — that list is fail-OPEN,
+so a registered task goes live on the next hourly wake unless the chart names it.
+
+**Trigger:** the queue stops being checked daily, or a row is found to have sat for weeks.
+
+---
+
+## 3 · The SAM → XRAS write direction
+
+Three things want it, and they are one capability:
+
+1. **Close abandoned requests.** Requests approved years ago and never pushed still
+   surface as people needing accounts (`NCAR0116` and friends, submitted 2015). That is
+   upstream data hygiene, not a SAM filter problem — the filters are working.
+2. **Buy lead time.** SAM sees these people at *approval*; they are knowable weeks
+   earlier, at submission. `SAM_TASKS_XRAS_SWEEP_STATUS` is already a knob.
+   ⚠️ Price it before building: 4,088 requests total vs 1,640 in-window, and a one-page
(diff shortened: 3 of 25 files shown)


benkirk:
## Gate 3 is closed — 13/13 byte-identical against `sam.hpc.ucar.edu`

Ran the parity check after opening this, so the last outstanding runbook precondition is now done rather than pending:

```
== xras == 13/13 checks passed (65.2s)
  ✓ people roster: 3,844,518 B identical      ✓ dates/requests single: 120 B identical
  ✓ people/{username}: 170 B identical        ✓ dates/requests multi:  306 B identical
  ✓ people/{username} 404: 58 B identical     ✓ requests/request/{SCSG0001,SCSG0002,UCIS0004}
  ✓ requests/user/benkirk: 6 masters          ✓ requests/role/{pi,co_pi,allocation_manager}
```

Two traps found on the way, both now recorded in the runbook (`8c242ba`), because each makes a **weaker run look like a stronger one**:

**The bare command silently tests less.** Without `--xras-user` the script samples the roster for a user with requests — and all eight it sampled had none. It warns, carries on, and reports `8/8 passed`, with `requests/request/*` exercised only against an unknown number and `dates/requests` **never probed**. `8/8` and `13/13` both print green. The working invocation is:

```bash
python utils/parity/check_legacy_apis.py --api xras \
       --new-base-url https://sam.hpc.ucar.edu --xras-user benkirk --timeout 120
```

**Legacy is slow enough to look like a failure.** It answers `requests/request/SCSG0001` in ~6.8s where the new stack takes ~0.67s, and dropped the connection outright on one run (`RemoteDisconnected`). Not a parity failure — probed by hand, both hosts return the same 30,296 bytes. Retry before investigating.

Worth noting for triage week: adding a resource-mapping row is the most likely 422 fix and it *moves GET response bytes*, so this gate will need re-running during the week rather than being done once.

benkirk:
## ✅ Deployed and validated — CIRRUS, 2026-08-20 19:04 MDT

`sha-82d8c00` from this branch (cirrus `f8933ee`). **Dispatch is live.**

### The flip

| | Before | After |
|---|---|---|
| Image | `sha-25ce500` | **`sha-82d8c00`** |
| `XRAS_ACTIONS_CAPTURE_ONLY`, both pods | `1` | **`0`** |
| The app's **parsed** value, `actions._capture_only()` | — | **`False` → dispatch LIVE** |

Checked the parsed value rather than only the env var: `"0"` reaching the pod and the app reading it as off are two different facts, and only the second one matters.

**The precondition held.** `xras_action_log` was **0 rows across every status** immediately before the flip, and still is after. ACCESS has not repointed, so nothing was stranded and no double-apply was possible — the check the runbook demands rather than assumes.

### Everything else that had to be true

- **All six handlers registered** — `add, adjust, extend, supplement, transfer, update`. Rules out the § 5 failure mode where an empty registry turns every action into a plausible-looking `manual` row.
- **`POST /api/xras/v1/actions` unauthenticated → 401, exactly 41 bytes**, legacy-identical. Cert valid to 2027-02-13, both SANs.
- **`--validate-mapping`: 13 keys offered by XRAS, 13/13 resolve.** The claim § 3.1 was rewritten around, confirmed against production. Six decommissioned mappings (Cheyenne, Yellowstone, HPSS, …) — harmless, reported.
- **`--validate-opportunities`, first production run: 43 map rows, 6 open opportunities, every one resolves, 0 dangling, exit 0.** The sweep had already backfilled 13 → 43, as #459 predicted.
- **`xras_sweep` ledger**: `succeeded`, `published: true`, **`publish_backend: redis`**, 21 pages, `budget_exhausted: false`, `unavailable_errors: 0`.
- **Accounts Needed has real content for Monday: 18** — 10 absent, 8 inactive, 9 placeholder.
- **`SAM_TASKS_XRAS_MAP_MAX = 20` reached the CronJob**, which does not inherit `webapp.env` — a genuine check, not a formality.
- 2/2 pods ready, 0 restarts, 0 ERROR lines since rollout.

### Two things worth recording

⚠️ **The mailer check has a trap.** `app.config.get('NOTIFY_ENABLED')` returns `None` in production and reads as fail-closed. It is not: `sam/notify/config.py` falls back to `os.environ` (the same seam as `xras_api/config.py`). The real accessor, `NotifyConfig.from_environment()`, reports `enabled=True`, `transport='smtp'`, `redirect_to=''` — live, to real recipients. Deriving it the obvious way gives the opposite answer.

✅ **§ 3.9's `map_budget_exhausted` semantics got confirmed by accident, on live data.** The 22:00 sweep wrote 20 of 30 unmapped opportunities and set the flag; the 23:00 run wrote the remaining 10 and the key is simply **absent** from that run's detail. "Missing means false" — written this afternoon, observed hours later.

### Still open

Only the ACCESS repoint. Until then `xras_action_log` stays at 0 and the **Accounts Needed** tab stays empty while **Pending Requests** is populated — the designed state, and now the one thing this PR makes sure nobody debugs on day one.

PR #458, opened 2026-08-20
XRAS outgoing queries: an account-creation worklist

benkirk:
## Summary

SAM's first **outbound** XRAS code. Everything under `src/sam/xras/` and
`src/webapp/api/xras/` is XRAS → SAM; this is the opposite direction, SAM
calling `https://api.xras.org/v1/…`.

It produces a dashboard card listing **the user accounts that must be created
or reactivated in SAM before an XRAS handoff can succeed**. Unreconciled ARC
placeholder identities are, by this repo's own measurements
(`src/sam/xras/handlers/new.py:24-27`), **55% of the legacy 70% failure
rate** — the single largest cause of handoff failure. Account creation is
manual, so the deliverable is a worklist, not automation.

Design and probe results: `docs/xras/outgoing/XRAS_OUTGOING_QUERIES.md`
(moved here from `docs/plans/` in the last commit; § 0 records what the build
learned).

## ⚠️ Read this before reviewing the card

There are **two feeds behind three tabs**, and they fail empty for opposite
reasons — which is the whole reason both exist:

- **Accounts Needed (Feed A)** reads `xras_action_log`, and renders
  legitimately **empty in production** until ACCESS is repointed at SAM. That
  table is at **0 rows** today. Empty is a designed state, not a broken one.
- **Pending Requests (Feed B)** reaches XRAS directly and therefore works
  *before* cutover. Verified live in production 2026-08-20: 22 requests,
  18 accounts needed.

*Unconfigured* is also a designed state: with `XRAS_OUTGOING_ENABLED` off,
Feed A still renders in full from the action log and Feed B renders empty,
with XRAS-sourced person detail marked unavailable.

## The commits

| # | Commit | What |
|---|---|---|
| 1 | `sam/integration/xras_api` | GET-only client + cache + probe script |
| 2 | `sam/queries/xras_accounts` | Two feeds → one classifier |
| 3 | `allocations dashboard` | The card, read-only |
| 4 | `sam-admin xras` | `--accounts`, `--person`, two-sided `--validate-mapping` |
| 5 | `scheduling: xras_sweep` | Enumerate-and-diff, **hourly on business hours** |
| 6 | `helm` | Fail-closed config, key via ExternalSecret |
| 7 | `docs` | File under `docs/xras/outgoing/`, settle the co-PI question |
| 8 | `tabs + Feed B` | Three-tab shell, shared date filter, request column/chips |
| 9 | `prod fixes` | `CACHE_REDIS_URL` + Redis NetworkPolicy for the task pods |

## Three things worth a reviewer's attention

**GET-only is structural, not conventional.** The same credential can create
requests, delete them, modify roles, and *merge one person into another*. So
the client has no verb method at all — `_get` is the only transport
primitive, and a test asserts no `post`/`put`/`patch`/`delete` callable
exists on the class. Likewise `XA-CONTEXT: report` is hardcoded, not a
parameter.

**PII is gated in the route, not the template.** Person detail (name, email,
organization, academic status, residence country) is assembled only for
`MANAGE_XRAS`; a `VIEW_XRAS` response never carries the bytes. A test asserts
the stubbed email, surname, organization and country are all *absent* from a
VIEW-only response body — not merely undrawn. `isReconciled` is deliberately
on the VIEW side: it is account state, and it is the signal an item is about
to close itself.

**`isAccountToBeCreated` is never the predicate.** XRAS sets it when a role
is created and never clears it. Validated against the unscrubbed corpus and
the dev database: **all 5 usernames carrying the flag were existing, active
SAM accounts** — 100% stale. Using it would have produced 5 false positives
and found none of the 9 real cases. It ships as a hint column with a
regression test pinning that a flagged active user does not appear.

## Verification

**Live probe against production** (`scripts/xras/probe_outgoing.py`, opt-in,
skips without a key):

- 13 resources, all carrying `resourceRepositoryKey`, reconciling **13/13**
  against `xras_resource_repository_key_resource` — zero unmapped, zero
  dangling. `sam-admin xras --validate-mapping` now proves this instead of
  assuming it.
- `reports/requests` paginates cleanly: 3 pages × 10, 30 distinct, zero overlap.
- Person objects arrive **inline** in every roster with `isReconciled` and
  `residenceCountry`.
- **Zero co-PIs across 64 sampled role entries**, confirming `/v1/types/roles`:
  the NCAR process declares exactly three role types, so a co-PI cannot appear
  on this wire. That closes a hedge carried in three places.
- 7 of 43 sampled role-people are `isReconciled: false` — the worklist has
  real content waiting.

**Tier-III, against the unscrubbed corpus + the dev database** (the only way
to validate the core predicate — scrubbed fixtures make "no `users` row"
trivially true for every row): 41/41 payloads parsed, 72 distinct real
usernames, **9 worklist rows — 4 absent, 5 inactive**. An existence-only
predicate would have found 4 of 9. Counts only; nothing derived from that
corpus is committed.

**Tests**: `7,196 passed, 42 skipped, 1 xfailed` across unit + integration +
api. 132 new tests. `bash helm/tests/test-cronjob-render.sh` passes, and its
new assertions were negative-tested (deleting `XRAS_API_USER` from the CronJob
exits 1; restoring it exits 0).

Three existing test rosters were updated — the cache-adapter list, the
cache-category set, and the audit window-form check. Each is a deliberate
enumeration that a new registration is *supposed* to break. The third had
been counting `name="days"` per *page* while its docstring described the
invariant per *form*; it now asserts what it means.

## Deployment notes

- **The OpenBao secret already exists** at `csg/xras-api-key`, property
  `token`, so `webapp.xrasApiCredentials.enabled: true` is safe on arrival.
- **Both levers ship ON**, and they are ONE decision: `XRAS_OUTGOING_ENABLED:
  "1"` with `xras_sweep` removed from `SAM_TASKS_DISABLED`. The task skips
  while the lever is off and the tab renders only what the task publishes, so
  a split state is just a dead tab; `helm/tests/test-cronjob-render.sh` fails
  if the two disagree. That list is fail-open — a registered task goes live on
  the next hourly wake unless the chart names it.
- `XRAS_API_KEY` is injected into **both** the Deployment and the CronJob.
  They share no env anchor; this is the same cross-referencing trap `NOTIFY_*`
  hit, and the render assertions are per-manifest for that reason.
- Nothing is written by any of this. No new table, no migration, no write path.

## Deferred, deliberately

`xras_account_event` (operator notes and dismissal — `XrasActivationEvent`
cannot carry it: `project_id` is NOT NULL and project-scoped, while this
worklist is username-keyed and a New request has no project yet) is the
immediate follow-up. Also deferred: the `opportunityId` → allocation-type map
(now sketched in `docs/plans/XRAS_OPPORTUNITY_ALLOCATION_TYPE.md`, designed to
be purely additive so ingestion never depends on the outbound API) and key
rotation. Each is recorded in § 14 of the design doc with what would change
the decision. **No longer deferred**: the sweep and the lever are both on,
and verified in production below.

## Verified in production, 2026-08-20

Deployed to CIRRUS and watched two sweep slots. **Two bugs only production
could show, both fixed in this branch:**

1. **The task pods had no `CACHE_REDIS_URL`.** `cronjob-tasks.yaml` renders
   `.Values.tasks.env` plus a hand-listed set and does *not* inherit
   `.Values.webapp.env` — the same trap `NOTIFY_*` hit. The sweep published
   into a per-worker cache that died with the pod, so the dashboard read
   nothing while the ledger said `published: true`.
2. **The Redis NetworkPolicy denied the task pods.**
   `samuel-redis-allow-webapp` admitted only `app: samuel`; task pods are
   `app: samuel-tasks`, so the connection timed out. Added the second peer.

`store_pending_worklist()` now returns *where* the snapshot landed
(`redis` / `local` / `disabled` / `full`) and `published` is true only for
`redis`, so this class of bug reports itself instead of claiming success.

| | 11:07 (pre-fix) | 12:07 (post-fix) |
|---|---|---|
| `publish_backend` | `local` | **`redis`** |
| `published` | `false` | **`true`** |
| duration | 59s | 47s |
| pages / in-window | 21 / 1640 of 4088 | 21 / 1640 of 4088 |
| accounts needed | 18 (10 absent, 8 inactive, 9 placeholder) | 18 (same) |

`unavailable_errors: 0` on both. Enumeration is stable run to run.

**Read-only Playwright pass against `sam.hpc.ucar.edu`**: the Pending Requests
tab renders the published snapshot, the shared date pills restyle *and* move
counts (7D → 2 rows, 90D → 9), zero console errors, zero 4xx/5xx. The one
unexplained 404 seen pre-publish did not reappear — it was an artifact of the
empty state.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -108,6 +108,29 @@ STATUS_API_KEY="YOUR_GENERATED_API_KEY_HERE"
 # SAM_XRAS_PASS=
 
 #-------------------------------------------------
+# XRAS Allocations API — SAM calling OUT to XRAS  (docs/xras/outgoing/)
+#
+# ⚠️ NOT the same thing as SAM_XRAS_USER / SAM_XRAS_PASS above. Those are XRAS's
+#    credential for calling SAM — a production WRITE credential in the inbound
+#    direction. This key is SAM's credential for calling api.xras.org.
+#
+# The SAME key can create requests, merge one person into another, and modify
+# roles. That is why the client is structurally GET-only: its one transport
+# primitive is an internal _get, and a test pins that no write verb exists.
+#
+# Production reads the key from OpenBao (csg/xras-api-key, property `token`)
+# via ExternalSecret. Locally, paste it below — and never commit a value.
+#
+# Fail-closed: with XRAS_OUTGOING_ENABLED=0 the dashboard card renders its
+# unconfigured state, the xras_sweep task records a skip, and the CLI degrades.
+# Nothing raises for lack of a key.
+XRAS_OUTGOING_ENABLED=0
+XRAS_API_KEY=
+XRAS_API_BASE=https://api.xras.org
+XRAS_ALLOCATIONS_PROCESS=NCAR
+XRAS_API_USER=arcguest
+
+#-------------------------------------------------
 # JupyterHub Live Statistics API
 # Used for real-time JupyterHub stats endpoints
 JUPYTERHUB_API_URL=https://jupyterhub.hpc.ucar.edu
--- a/docs/README-k8s.md
+++ b/docs/README-k8s.md
@@ -247,7 +247,7 @@ Scheduled tasks, by environment:
 |---|---|---|
 | Local Docker Compose (`webdev`) | n/a — no chart | Run by hand: `sam-admin tasks --run-due` |
 | Local k8s (Docker Desktop) | `false` | Nothing should silently DELETE local data |
-| CIRRUS k8s (this chart) | `true`, kill-switched | Staged enable: only `cleanup_status_snapshots` runs. `SAM_TASKS_DISABLED=deactivate_expired_projects,expiration_notices,xras_notices` |
+| CIRRUS k8s (this chart) | `true`, kill-switched | Staged enable: only `cleanup_status_snapshots` runs. `SAM_TASKS_DISABLED=deactivate_expired_projects,expiration_notices,xras_notices,xras_sweep` |
 
 When the per-environment Entra app strategy is adopted (separate `sam-production`
 and `sam-staging` Entra apps), only the OpenBao / SSM values change — the chart
@@ -329,7 +329,8 @@ of task names to skip, flippable in `values.yaml` with no code deploy. It ships
 until it has been reviewed on its own, so the dispatcher wakes hourly and the
 untried task writes a `skipped` row instead of running. Today only
 `cleanup_status_snapshots` is live; `deactivate_expired_projects`,
-`expiration_notices` and `xras_notices` are switched off. Enabling one is a
+`expiration_notices`, `xras_notices` and `xras_sweep` are switched off.
+Enabling one is a
 separate, reviewable one-line commit.
 
 ⚠️ **It is an enumeration, and it is fail-OPEN.** `disabled_tasks()` in
@@ -368,7 +369,7 @@ that needs no task to actually do anything:
 ```bash
 helm upgrade --install samuel ./helm -f helm/values.yaml -f helm/values-local.yaml \
   -n samuel-dev --set tasks.enabled=true --set tasks.schedule='*/5 * * * *' \
-  --set 'tasks.env.SAM_TASKS_DISABLED=cleanup_status_snapshots\,deactivate_expired_projects\,expiration_notices\,xras_notices'
+  --set 'tasks.env.SAM_TASKS_DISABLED=cleanup_status_snapshots\,deactivate_expired_projects\,expiration_notices\,xras_notices\,xras_sweep'
 ```
 
 ⚠️ `--set` **replaces** the list rather than adding to it, and the commas need
--- a/docs/plans/XRAS_OPPORTUNITY_ALLOCATION_TYPE.md
+++ b/docs/plans/XRAS_OPPORTUNITY_ALLOCATION_TYPE.md
@@ -1,0 +1,332 @@
+# `opportunityId` → allocation type — a mapping table, built additively
+
+**Status:** designed, **not implemented**. Sketched 2026-08-20 with the operator.
+**Base:** stacks on **PR #458** (`probing_xras` → `staging`, 19 commits, all CI
+green). Branch from that head; do not branch from `staging`.
+**Origin:** the deferred § 8.2 item in
+`docs/xras/outgoing/XRAS_OUTGOING_QUERIES.md`, reshaped around one hard
+constraint (§ 2).
+
+This is a handoff document. It assumes **no prior context** — an implementation
+session should be able to start cold from here.
+
+---
+
+## 1. What you need to know first
+
+XRAS is the ACCESS allocations broker. It **pushes** allocation decisions into
+SAM at `POST /api/xras/v1/actions`; `src/sam/xras/` is the handler stack that
+applies them. PR #458 added the opposite direction — a **read-only, GET-only**
+client at `src/sam/integration/xras_api/` that calls `https://api.xras.org/v1/…`
+— plus a dashboard worklist of accounts that must exist before a handoff can
+succeed, and an hourly `xras_sweep` task.
+
+None of that is required reading for this work. What matters here is one field
+on the **inbound** wire that SAM currently ignores.
+
+### The ladder, as it stands
+
+`resolve_allocation_type(session, action, errs)` — `src/sam/xras/extractors.py:323`
+— decides which `allocation_type` row a project gets. It runs an
+**11-strategy chain** of string matching over the wire fields `allocationType`,
+`opportunityName` and `requestTitle`, producing a
+`SelectionParms(panel: str, allocation_type: str)`, then joins:
+
+```python
+row = (session.query(AllocationType)
+       .join(Panel, AllocationType.panel_id == Panel.panel_id)
+       .filter(Panel.panel_name == parms.panel)
+       .filter(AllocationType.allocation_type == parms.allocation_type).first())
+```
+
+Facts worth having in front of you:
+
+| | |
+|---|---|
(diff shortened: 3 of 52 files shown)


benkirk:
## Local smoke — four real bugs, all fixed in this branch

Seeded the 41-payload unscrubbed corpus into the dev database and drove the
card in a browser. Every finding below was invisible to the in-tree fixtures.

**1. The worklist was regime-dependent.** `WORKLIST_STATUSES` excluded
`processed` on the reasoning that a succeeded action's roster must have
resolved. It doesn't: `resolve_roster` treats a missing or inactive *member*
as a warning, so the handler skips them and succeeds while the account is
still needed. Same 41 payloads gave **9 rows as `received`, 4 once dispatched**
— hiding four inactive users and one absent Allocation Manager. That is the
exact property §1 says the worklist must not have.

**2. A case-mismatched username was a false positive.** `users.username` is
`utf8mb3_general_ci`, so the SQL matched but the Python dict underneath was
case-sensitive — reporting "create this account" for an account that exists
and is active. The anonymizer emits lowercase `user_<hex>`, so no scrubbed
fixture could ever reach this.

**3. `isReconciled` is not a closure signal.** The design says re-poll it and
the item closes itself. Live: **9 of 9 rows reconciled in XRAS, every one still
needing a SAM account.** Reconciliation is XRAS linking a placeholder to a real
identity; it says nothing about SAM. Renamed to "XRAS identity"
(identified/unidentified), with the *unidentified* case flagged as the harder
one. The genuine closure signal was already free — classification is
current-state, so a row leaves when its `users` row appears.

**4. The sweep reported a census, not a queue.** Against the live process it
found 2,180 "accounts needed" over 4,088 requests, 2,149 merely inactive.

| Feed B scope | rows |
|---|---|
| every Approved request | 2,180 |
| + period-of-performance window | 542 |
| + not-yet-pushed only | **21** — 10 of 11 absent are ARC placeholders |

The window was specified in §7.4 and I hadn't implemented it. The second filter
is the one that matters: a request whose project already exists **has already
had its handoff**, so its roster's inactive members are ordinary attrition.
Role was tried first and rejected — PI/AM-only still left 344 inactive.

Live result now: `1640/4088 in-window, 40 pending push, 21 accounts needed`.

## Also

- The browser caught a header badge reading "3 unreconciled" directly above
  three rows each reading "identified" — two independent facts conflated.
- Three Feed-A tests asserted on a globally-empty action log, racing the card
  tests' committed fixture across xdist workers. Now scoped to their own rows.
- New `e2e/test_xras_accounts_card.py`: 7 browser tests, guarded so an empty
  stack doesn't redden the build.

7,227 passing across unit + integration + api; helm render green.

## Still open for a follow-up

**Feed B has no dashboard surface.** The sweep's 21 accounts and 40 pending
pushes land only in `TaskResult.detail`; the card is Feed A only, by the
design's own deferral ("Feed B via task detail, table later"). Now that the
numbers are a real queue rather than a census, surfacing them on the card is
the obvious next PR — and the argument for enabling `xras_sweep` rests on it.

benkirk:
## Feed B surfaced + UX round (commits 12-17)

The sweep's findings now have a home on the dashboard, and the two worklists
are tabbed.

**Transport.** Feed B can't be computed in a request — the enumeration is 21
pages and 60-90s. So `xras_sweep` became a producer and the tab a consumer,
with a new `xras_pending` cache bucket as the mailbox. The TTL is 24h rather
than tuned to the hourly cadence: the real gap is 17:00→08:00, and a 4h TTL
would blank the tab every morning until the first sweep — exactly when
someone looks.

**The sweep is now hourly (business hours) and enabled**, with
`XRAS_OUTGOING_ENABLED` flipped to `"1"`. Those two are one decision and the
chart now enforces it — the task skips while the lever is off, and the tab
renders only what the task publishes, so enabling one without the other gives
a permanently empty tab and a ledger full of `skipped` with nothing failing to
say so. `test-cronjob-render.sh` fails if they disagree.

**Tabs.** Three panes over one window control. The control is rendered once,
in the shell, because `window_pills` binds its date pair with `form=` and
rendering it per tab would put three same-named pairs in one form — the
RadioNodeList trap. Reaching three panes needs no JavaScript: the pills' action
ends in `htmx.trigger(form, 'submit')`, which bubbles, so each pane listens for
`submit from:#xras-window-filters`.

**UX:** row icons dropped (placeholder is now a text badge); Request is both a
column and a chip on both worklist tabs.

### The pills bug, and the two behind it

Reported from the browser as "clicking 7D does nothing". Three distinct causes:

1. **The pill never restyled.** `window_pills` marks the active pill
   server-side; that came free while the pills lived inside each fragment.
   Moving them to the shell put them outside every swap target. They're now
   their own swap target.
2. **The pills were inert on Pending Requests** — that route never read the
   window at all.
3. **Fixing #2 naively hid the wrong rows.** Filtering Feed B on
   period-of-performance keeps every row at every width (pending allocations
   end a year out), so the pill still looked dead. The right axis is
   `submitDate` — the same question Feed A asks with `received_time`. But the
   live data then showed the sweep finding **21** accounts while 90D displayed
   **12**, and on that tab an *older* unpushed request is the more urgent one.
   A recency filter silently dropping those inverts the priority the tab
   exists to surface. The header now reads
   "10 of 21 account(s) needed · 11 outside the window".

### Verification

7,230 unit+integration+api, 11 browser tests, helm render. Verified in a
browser across all three tabs: pills restyle and move row counts 3/5/8
(Accounts) and 2/10/12 (Pending Requests). Feed B rendered 21 rows from a real
sweep, 7 placeholders `unidentified` — a distinction Feed A never surfaced.

One self-inflicted CI failure worth recording: an unguarded `thead.inner_text()`
in the new browser test waits out a 30s timeout on CI's empty stack rather than
failing fast. Fixed, and verified by reproducing the empty-stack condition
locally rather than assuming.

PR #459, opened 2026-08-20
XRAS: an opportunityId -> allocation-type map, with auto-detection

benkirk:
## Summary

XRAS's `opportunityId` is on **41/41** observed inbound payloads and read by
nothing. This makes it decide the allocation type, via a local mapping table
read at ingest, with the eleven-strategy free-text ladder kept as the fallback —
and then teaches `xras_sweep` to fill that table in by itself.

Design: [`docs/xras/outgoing/XRAS_OPPORTUNITY_ALLOCATION_TYPE.md`](docs/xras/outgoing/XRAS_OPPORTUNITY_ALLOCATION_TYPE.md)
(§ 0 and § 8.5 record what the build learned).

## The failure this disarms

`resolve_allocation_type` picks a `(panel, type)` pair from the ladder. The
twelve pairs it can produce **never name `UW`, `WRAP` or `LCAP`** — the whole of
facility 4 is unreachable through it.

So when Wyoming eventually submits, a `Small` request resolves to panel
`UNIV USS` and the join **succeeds**, because that is a valid row. Nothing
fails. The only symptom is a WNA project holding a UNIV projcode, since
`handlers/new.py:175` draws the series from `allocation_type.panel.facility_id`.
Projcodes are not undoable.

Every other mapping gap here shouts — an unmapped `resourceRepositoryKey` 422s
the action. This one is silent.

## ⚠️ Additive by construction

**An empty table reproduces the ladder exactly.** Ingestion reads a local table
and never calls out, so an inbound action never depends on
`XRAS_OUTGOING_ENABLED` or on `api.xras.org` being up. The first tests written
say so: with the table cleared, all 41 corpus payloads resolve to the pair they
resolve to today and five distinct strategies are still exercised.

## The commits

| # | Commit | What |
|---|---|---|
| 1-2 | `sam/integration/xras`, `sam/xras` | The table, the shared lookup, **both** call sites |
| 3 | `scheduling` | The sweep counts unmapped ids — no extra API calls |
| 4-5 | `docs`, `db` | Filed under `docs/xras/`, DDL recorded, snapshot |
| 6 | `helm/tests` | An unrelated flake that blocked this PR (below) |
| 7-12 | the auto-detection round | Constant, provenance, client, the agree rule, docs, snapshot |

## Part 1 — the map

**The design's own premise was never checked, and it holds.** Nothing asked
whether `opportunityId` is *single-valued*. It is: nine ids, five pairs, one
each across all 41 payloads. Now pinned.

**The sharp edge is `auth_at_panel_meeting`, not `resolve_allocation_type`.** It
re-derives the pair *independently* — one sets `project.allocation_type_id`, the
other sets `auth_at_panel_mtg` on **allocation_transaction** rows. Wiring only
the first would write inconsistent rows with nothing raised and nothing logged.
**Only its first arm changed**; the second reads the stored project type, which
the mapped resolver is what wrote.

**`allocation_type.panel_id` is NULLABLE**, so the obvious
`row.allocation_type.panel.panel_name` raises `AttributeError` mid-dispatch. A
null panel is a miss and falls through.

**`Small` is the only *silent* case.** `Education` is also a two-panel name but
is absent from the ladder's twelve type names, so a WNA `Education` request 422s
loudly.

## Part 2 — auto-detection

The operator posts ~4 opportunities a year (University Large ×2, NSC ×2), each
previously a hand-written INSERT. The sweep now writes the row itself.

**The full enumeration changed the design.** 21 pages, 4,088 requests: **42
distinct opportunities**, not the 9 in the corpus. They collapse to **eight**
stable `(allocationTypeId, primary panelId)` pairs — every University Large
since 2021 shares one, every NSC request another. So the reference half is an
eight-entry **constant**, test-asserted to name real `allocation_type` rows.

⚠️ **And it disproved this document's own § 3.** XRAS is *not* authoritative
about SAM: it files the unsponsored family under `Educational` (the same id as
Classroom), and gives `NCAR - ASD Opportunity` NSC's own type **and** panel.
Both change the **facility**. Four rows are therefore `manual`.

**So the rule is: two derivations must agree** — the constant and the ladder —
or the row is withheld and reported. That rule:

- caught **all four** bad cases without knowing about any of them, including two
  the design never anticipated, found by the rule rather than by hand;
- **withholds the first Wyoming opportunity**, which is also the alerting the
  follow-up asked for;
- makes an error in the constant **self-limiting**;
- held under a **partially-loaded database**, observed by accident during a
  snapshot rebuild — 41 candidates withheld, nothing written.

One row proves the key must be the **pair**: a 2018 NSC request is filed under
the *Large* type id on the NSC panel. Type id alone would have called it CHAP.

## ⚠️ A bug no unit test could catch

The new helper was first placed between `@task(...)` and `def xras_sweep`, which
registered **the helper** as the task body. The name is a decorator argument so
registration looked correct, and every unit test passed because they call
`mod.xras_sweep` directly. It failed only at dispatch — found by running the
task for real against the live API. There is now a guard, negative-tested.

## Verification

**Tests**: `7,319 passed, 42 skipped, 1 xfailed`. ~85 new. Run against a
`mysql-test` volume destroyed and restored from the committed blob alone.

**Every new guard was negative-tested**: reverting `auth_at_panel_meeting` fails
exactly one test; disabling the map lookup fails exactly three; reintroducing the
decorator bug fails exactly one.

**Live rehearsal** against the production XRAS API (`--dry-run`, which the runner
makes a true rehearsal by rolling back): **42 seen, 29 unmapped, 29 agreeing, 0
needing review, 0 unknown** — written 20 then 9 across two hourly runs.

**Production cross-checked** independently as `hpc-reader`: 13 rows, PK and FK
present, zero dangling, em-dash stored as `E28094` with no mojibake.

## Deployment notes

- **DDL and seed already applied to production** (2026-08-20), recorded in
  `XRAS_CUTOVER_RUNBOOK.md` § 2c. Two statements: the table, then
  `ADD COLUMN source`.
- **No config, no chart change, no new lever.** `xras_sweep` and
  `XRAS_OUTGOING_ENABLED` are already on from #458.
- The first sweep after deploy backfills 29 rows over two runs, all tagged
  `source='task:xras_sweep'` and all corroborated. `SAM_TASKS_XRAS_MAP_MAX`
  (default 20) bounds it.

## One unrelated commit

`helm/tests: the assert helpers could report either answer wrongly` — both render
tests ran `printf ... | grep -q` under `pipefail`, so a SIGPIPE on an early match
inverted the result. `assert_contains` gave a false FAIL (which blocked this PR);
**`assert_not_contains` gave a false PASS**, which is the serious half, since that
is how those tests prove a value is *absent*. Reproduced deterministically and
fixed with a herestring. Happy to split it out if preferred — it touches nothing
else.

## Deferred

`sam-admin xras --validate-opportunities`. The decision function and the client
method both exist, so it is CLI wiring only.


benkirk:
## ✅ Verified live in production

Deployed to CIRRUS as `sha-25ce500` and exercised against the real XRAS API. The
acceptance test was deliberately a *new* event rather than a synthetic one: a new
opportunity was posted in XRAS, and the sweep was left to find it unattended.

**It did.**

```
535388  CHAP / CHAP  fac=2  source=task:xras_sweep
        "Large Allocation (University) - Fall 2026"
```

No SQL was written by hand.

### The two runs

| slot | result |
|---|---|
| 22:07Z | `open=6 seen=43 unmapped=30 written=20 cap=True review=[] unknown=[]` |
| 23:07Z | `unmapped=10 written=10`, cap not hit |

**535388 landed in the first batch**, which is the newest-first cap ordering
earning itself: it is the highest id of 30 candidates against a cap of 20, so
under lowest-id-first it would have queued behind opportunities closed since 2018
and waited an extra slot.

### Final state

| | |
|---|---|
| Rows | **13 `manual` + 30 `task:xras_sweep` = 43** |
| Needing review / unknown pair | **0 / 0** across both runs |
| The four `manual` rows | untouched — `530296/530315/530900 → 4`, `531461 → 17` |
| Dangling or panel-less | **0** |

Auto-written distribution matches the eight-pair constant exactly: CHAP 16,
NCAR-ARP/NSC 7, UNIV USS/Small 5, Classroom 1, External Project 1.

### What the live run proved that the suite could not

**The open-list discovery was load-bearing, and nearly shipped missing.** The
sweep originally discovered opportunity ids only from `reports/requests`, which
by construction cannot mention an opportunity that has no requests. 535388 had
**zero**, confirmed via `/v1/reports/opportunity_requests/535388`, so it would
have been invisible until its first request was *approved* — potentially weeks,
and precisely the lead time this map exists to buy. `seen=43` rather than 42 is
that one opportunity.

**The agree-only rule ran clean on 30 real writes.** Zero review, zero unknown —
because the four cases where XRAS is wrong about SAM are pinned `manual`, and
the sweep only ever inserts where no row exists.

**Provenance is a working audit trail.** Every automated row is identifiable, and
the whole set is reversible in one statement if it ever misbehaves:

```sql
DELETE FROM xras_opportunity_allocation_type WHERE source = 'task:xras_sweep';
```

### Deployment checks that passed

- **Both** `deploy/samuel` and `cronjob/samuel-tasks` moved to `sha-25ce500` —
  they share one image key, and a webapp-only roll would have left the sweep on
  old code with nothing erroring.
- `XRAS_OUTGOING_ENABLED=1` and `XRAS_API_KEY` present **on the CronJob**, not
  merely on the Deployment (task pods inherit nothing from `webapp.env`).
- `xras_sweep` absent from `SAM_TASKS_DISABLED`, which is fail-open.
- Production schema cross-checked independently as `hpc-reader` before and after:
  PK, FK and index present, em-dash stored as `E28094` with no mojibake.

### One honest caveat

The deploy-watcher used for this went **silent** rather than reporting an error
when the VPN dropped mid-window — indistinguishable from "nothing happened",
though both slots had in fact run correctly. That is a flaw in the observation
tooling, not in this change, but worth stating: the evidence above was gathered
by direct inspection afterwards, not by trusting the watcher's silence.

